PR #24, opened 2021-05-06
chore: enable prettier

MrArnoldPalmer:
enables prettier for code formatting via `eslint --fix`.


.projenrc.js
```diff
@@ -1,33 +1,50 @@
...
 
 // npm tarball will only include the contents of the "build"
 // directory, which is the output of our static website.
-project.npmignore?.addPatterns('!/build');
-project.npmignore?.addPatterns('/public');
+project.npmignore?.addPatterns("!/build");
+project.npmignore?.addPatterns("/public");
+
+// setup linting for create-react-app specific tools
+project.eslint.addRules({
```
eladb, line 39: Should this go in projen?

eladb (approved):

.projenrc.js
MrArnoldPalmer: It's kinda funky because the current eslint rules aren't really friendly to the react template and requires a lot of overrides. Prettier removes some of these though, but I'm working on a separate PR to make the default rules a bit more composable and project specific where needed.

Issue #3, opened 2021-04-21
Design React App


RomainMuller:
Design the React app component model, in order to streamline implementation.

Determine how routing is going to be implemented and finalise the URL scheme used to serve pages to the customer.

Finalise the CloudFront Distribution configuration.

PR #34, opened 2021-05-13
feat: package routing

MrArnoldPalmer:
Initial routes for package pages with scope, name, and version path
params.

fix: #3

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -34,6 +34,10 @@
       "type": "build"
     },
     {
+      "name": "@types/react-router-dom",
+      "type": "build"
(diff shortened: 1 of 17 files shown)


eladb (approved):
Looks good. 

Can you add the to the PR description the list of supported routes?

eladb:
Does this "fix" #3 or just related to it? 

I think we need some general design discussion on how we model the app that's not covered here.

MrArnoldPalmer:
There are some other tasks related to #3 but they are basically just making decisions. I'm okay with having it closed from this.

PR #62, opened 2021-05-25
chore: add storybook

MrArnoldPalmer:
Adds storybook for component development and testing. Resynth projen
project to add necessary dependencies and scripts.


.projen/tasks.json
```diff
@@ -98,35 +98,27 @@
     },
     "upgrade-dependencies": {
       "name": "upgrade-dependencies",
-      "description": "upgrade dependencies",
```
eladb, line 101: Looks like you downgraded projen for some reason... Are you running `npx projen` or globally installed?

.storybook/main.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
eladb, line 1: In the future, would it make sense to generate this from .projenrc.js?

.projen/tasks.json
MrArnoldPalmer: Whoops. Usually I yarn install and then 'yarn projen' to keep it local. 'npx projen' has given me some issues in the past. Will fix

.storybook/main.js
MrArnoldPalmer: It could, this was generated by the storybook cli so eventually it might be something to include in projen as an option for react projects.

.projen/tasks.json
```diff
@@ -198,6 +198,22 @@
...
+    },
+    "storybook": {
+      "name": "storybook",
+      "steps": [
+        {
+          "exec": "start-storybook -p 6006 -s public"
+        }
+      ]
+    },
+    "build-storybook": {
```
eladb, line 210: 1. Can you add descriptions to the tasks?
2. Maybe call this one `storybook:build` (no strong opinion)

.projenrc.js
```diff
@@ -44,11 +53,32 @@ project.eslint.addRules({
...
+        "**/*.test.ts",
+        "**/*.stories.*",
+      ],
       optionalDependencies: false,
       peerDependencies: true,
     },
   ],
 });
 
+// Add tasks and config for storybook
```
eladb, line 68: Could you package all the storybook related stuff under a function instead of spread across the file?

(Would be useful to extract to a projen component later)

eladb (commented):
Minor

.projenrc.js
```diff
@@ -28,6 +28,44 @@ const project = new web.ReactTypeScriptProject({
...
+    "babel-loader@8.1.0"
+  );
+
+  // Add tasks and config for storybook
+  project.addTask("storybook", {
+    exec: "start-storybook -p 6006 -s public",
+    description: "run local storybook server",
+  });
+
+  project.addTask("build:storybook", {
```
eladb, line 48: Should this part of the main package compile?

```ts
const build = project.addTask(..);
project.compileTask.spawn(build);
```

I'd also call this `storybook:build` instead of `build:storybook`

.projenrc.js
```diff
@@ -28,11 +28,53 @@ const project = new web.ReactTypeScriptProject({
...
+  project.gitignore?.addPatterns("/storybook-static/");
+})();
+
 // synthesize project files before build
 // see https://github.com/projen/projen/issues/754
-const build = project.tasks.tryFind("build");
-build.prependExec("npx projen");
-build.spawn(project.packageTask);
+const buildTask = project.tasks.tryFind("build");
+buildTask.prependSpawn(project.compileTask);
```
eladb, line 75: I don't think this is needed. The compile task is part of build already

RomainMuller: Nope, it is not. For some reason, it is there at the beginning of the workflow, but disappears later.... And I don't have time to investigate that right now.

iliapolo (approved):

Issue #46, opened 2021-05-23
Implement: Package page - package data - content area (python)


iliapolo:
This is the main content of the package page. It is a markdown formatted page that includes both the user defined `README.md`, as well as a generated API reference doc based on the jsii assembly file.

We should leverage `jsii-reflect` to create a generic function that accepts a `jsii` assembly file and outputs the necessary documentation string. The output should be such that it provides the necessary information for a user to be able to instantiate  and use any methods of any construct. 

This means providing documentation on every argument that is expected to be passed in by the user, including enums, constructs, and interfaces - not just construct props.

This function will later be used by a react component that does the actual rendering.

Should only support python for now.

## Notes

- No need to transliterate code snippets in the user defined `README.md` file.
- Requires some code changes to `jsii-reflect` to make it browser compatible.

PR #68, opened 2021-05-30
feat: Package documentation in python

iliapolo:
Render a full documentation page for a specific package.

The PR is pretty massive, though many of the added files are simply wrapper classes that don't contain much logic.
To facilitate the review, here is the main structure of the code.

### view

This directory represents the elements that we want to show on the page:

- Constructs
- Interfaces
- Enums
- ...

Each of these elements has nested elements, so basically every such element is located in its own file.
These components are language agnostic (hopefully).

### transpile

This directory defines the necessary language specific information in the form of a `Transpile` interface with language specific implementations (currently only python).

### render

This directory includes rendering methods. Currently only markdown.

-----------

I acknowledge this might not necessarily hold as we introduce new languages, but hopefully gives us a good framework to start from.

Fixes https://github.com/cdklabs/construct-hub-webapp/issues/46


.projenrc.js
```diff
@@ -80,14 +83,21 @@ const project = new web.ReactTypeScriptProject({
 // synthesize project files before build
 // see https://github.com/projen/projen/issues/754
 const buildTask = project.tasks.tryFind("build");
+buildTask.spawn(project.packageTask);
+buildTask.prependSpawn(project.testTask);
```
iliapolo, line 87: Am I missing something or did we not run tests as part of the build?

.projenrc.js
```diff
@@ -105,3 +115,78 @@ project.eslint.addRules({
...
+
+  if (fs.existsSync(assemblyPath)) {
+    console.log(`Assembly for ${package} already exists locally`);
+    return;
+  }
+
+  console.log(`Fetching assembly for ${package}`);
+  fs.mkdirSync(path.dirname(assemblyPath), { recursive: true });
+  exec(
+    `cd ${tempDir} && npm v ${packageName}@${packageVersion} dist.tarball | xargs curl | tar -xz`,
```
iliapolo, line 179: `dist.tarball` is a special string that gives you the tarball URL from `npm view`.

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,318 @@
...
+        parameters.push(p);
+      } else {
+        // struct parameters are expanded to the individual struct properties
+        const struct = p.parentType.system.findInterface(p.type.fqn!);
+        for (const property of struct.allProperties) {
+          const parameter = propertyToParameter(callable, property);
+          parameters.push(parameter);
+        }
+      }
+    }
```
iliapolo, line 155: This expands every struct parameter, which is probably wrong. I'm not sure how to identify just the one I need. Can I rely on the order and take the last? @RomainMuller ?

src/api/docgen/view/parameter.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+function findParameter(): reflect.Parameter {
+  for (const klass of assembly.system.classes) {
+    for (const method of klass.ownMethods) {
+      if (method.parameters.length > 0) {
+        return method.parameters[0];
+      }
+    }
+  }
+  throw new Error("Assembly does not contain a paramter");
```
addihorowitz, line 23: typo

src/setupTests.ts
```diff
@@ -3,3 +3,42 @@
...
+(global as any).assembly = createAssembly();
+
+function collectAssebmlies(p: string, ts: reflect.TypeSystem) {
+  const stat = fs.statSync(p);
+
+  if (stat.isDirectory()) {
+    fs.readdirSync(p).map((f) => collectAssebmlies(path.join(p, f), ts));
+  } else {
+    if (!p.endsWith("jsii.json")) {
+      throw new Error(
```
addihorowitz, line 35: Does it actually requires throwing an exception? Shouldn't we just log it and move on?

src/setupTests.ts
```diff
@@ -3,3 +3,42 @@
...
+        `Unexpected file in packages: '${p}' (should contain only jsii.json files)`
+      );
+    }
+    const assembly = JSON.parse(
+      fs.readFileSync(p, { encoding: "utf8" })
```
addihorowitz, line 40: We should  have validation method here. To make sure the JSII file has the correct structure, to sanitize the text for security reasons etc.

addihorowitz (commented):
Hi Eli, good work. I added some comments and questions (from a non-React-expert perspective :)). Please review

.projenrc.js
```diff
@@ -105,3 +115,78 @@ project.eslint.addRules({
...
+  task.exec(`node ${scriptPath}`);
+}
+
+/**
+ * Fetch an assembly by name and version, accepted via env variable for easier invocation.
+ * This fetches all the necessary assemblies into the packages path, which can later be used to instantiate
+ * a jsii type system for the specific package.
+ */
+function fetchAssembly(packagesPath) {
+  const packageName = process.env.PACKAGE_NAME;
```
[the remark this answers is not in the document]
iliapolo: Outdated. They are read as environment variables in the main script and passed to the function as arguments. 
The reason the main script extracts them from the environment is to make it easier to invoke via `projen dev:fetch-assemblies`, which doesn't pass through arguments properly.

.projenrc.js
```diff
@@ -105,3 +115,78 @@ project.eslint.addRules({
...
+    packageVersion = packageVersion.substring(1, packageVersion.length);
+  }
+
+  const output = `${packagesPath}/${packageName}@${packageVersion}/jsii.json`;
+
+  const tempDir = fs.mkdtempSync(`${os.tmpdir()}/assembly`);
+  const assemblyPath = `${__dirname}/../${output}`;
+  const package = `${packageName}@${packageVersion}`;
+
+  if (fs.existsSync(assemblyPath)) {
```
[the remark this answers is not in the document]
iliapolo: This specifically is just a dev script that always fetches assemblies by a specific version. In the app itself assemblies are also always fetched by a specific version, where `latest` will somehow need to be resolved to the latest available concrete version.

eladb:
@iliapolo Any chance you can share an example of the rendered output for a well known module?

scripts/fetch-assemblies.js
```diff
@@ -0,0 +1,55 @@
+// ~~ Generated by projen. To modify, edit .projenrc.js and run "npx projen".
```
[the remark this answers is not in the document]
iliapolo: Its not used in the app - this is just a development script generated by projen for convenience. Its not generated at build so we can't `.gitignore` it, nor should we really, its just a source file I happened to use projen to author :)

src/Package/index.tsx
```diff
@@ -13,25 +17,20 @@ interface PathParams {
...
-  readme?: string;
+  assembly?: reflect.Assembly;
   error?: string;
 }
 
 export function Package() {
   const { name, scope, version }: PathParams = useParams();
-  const [{ assembly, readme }, setState] = useState<PackageDocsHomeState>({});
+  const [{ assembly }, setState] = useState<PackageDocsHomeState>({});
+  const q = parseSearch(useLocation().search);
```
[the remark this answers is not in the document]
iliapolo: I don't believe `useLocation` can fail, it just returns the current URL, and `.search` will return an empty string if no query parameters are specified, which is handled by `parseSearch`.

iliapolo:
> @iliapolo Any chance you can share an example of the rendered output for a well known module?

@eladb https://gist.github.com/iliapolo/8eac278b9288db7cc8b45a78ef303627

src/Package/index.tsx
```diff
@@ -44,17 +43,80 @@ export function Package() {
...
+  scope?: string
+): Promise<spec.Assembly> {
+  if (version.startsWith("^")) {
+    version = version.substring(1, version.length);
+  }
+
+  // e.g https://awscdk.io/packages/@aws-cdk/alexa-ask@1.106.0/jsii.json
+  const assemblyPath = `${getAssetsPath(name, version, scope)}/jsii.json`;
+  const response = await fetch(assemblyPath);
+  return response.json();
```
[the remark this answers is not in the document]
iliapolo: Added a validation that the request succeeded and throw if not 👍

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,190 @@
...
+  public split() {
+    this.lines("---");
+    this.lines("");
+  }
+
+  public section(section: Markdown) {
+    this._sections.push(section);
+  }
+
+  public render(headerSize: number = 0): string {
```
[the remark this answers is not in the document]
iliapolo: Added a comment in the code.

src/api/docgen/view/documentation.ts
```diff
@@ -0,0 +1,114 @@
...
+    return documentation;
+  }
+
+  private findSubmodule(submoduleName: string): reflect.Submodule {
+    const submodules = this.assembly.submodules.filter(
+      (s) => s.name === submoduleName
+    );
+
+    if (submodules.length === 0) {
+      throw new Error(`Submodule ${submoduleName} not found`);
```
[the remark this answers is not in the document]
iliapolo: An exception is more appropriate because there is nothing we can do if the submodule doesn't exist, this component shouldn't exist and shouldn't be initialized. 

Maybe we should introduce a specific exception type so that callers can easily identify such situations, but this can be added easily when such a use-case arises (probably soon with proper UI error handling).

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,318 @@
...
+    const type = this.ts.findFqn(fqn);
+    const submodule = this.findSubmodule(type);
+
+    const targets = submodule ? submodule.targets : type.assembly.targets;
+
+    if (!targets) {
+      throw new Error(`Unable to find 'targets' for fqn ${type.fqn}`);
+    }
+
+    if (!targets.python) {
```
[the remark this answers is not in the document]
iliapolo: Yes, if there is no python target at this point something is horribly wrong and the docs will be incorrect.

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,318 @@
...
+    // if the type is in a submodule, the submodule name is the first
+    // part of the namespace. we construct the full submodule fqn and seach for it.
+    const submoduleFqn = `${type.assembly.name}.${
+      type.namespace.split(".")[0]
+    }`;
+    const submodules = type.assembly.submodules.filter(
+      (s) => s.fqn === submoduleFqn
+    );
+
+    if (submodules.length > 1) {
```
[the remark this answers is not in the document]
iliapolo: Added a comment. This can't happen. Its essentially just a side-effect of converting a dictionary of submodules (as stored in the spec) to an array (as exposed here).

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,318 @@
...
+  }
+  return `${target}(${formatInputs(inputs, 1 + target.length)})`;
+}
+
+function formatImport(type: transpile.TranspiledType) {
+  return `import ${type.moduleFqn}`;
+}
+
+function formatSignature(name: string, inputs: string[]) {
+  return `def ${name}(${formatInputs(inputs, 3 + 1 + 1 + name.length)})`;
```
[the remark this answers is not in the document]
iliapolo: Added comment

src/api/docgen/view/__snapshots__/class.test.ts.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
iliapolo: Yeap, this is what the tests assert against.

src/api/docgen/view/__snapshots__/enum.test.ts.snap
```diff
@@ -0,0 +1,22 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
iliapolo: https://github.com/cdklabs/construct-hub-webapp/pull/68#discussion_r642105279

src/api/docgen/view/api-reference.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.structs = new Structs(transpile, interfaces);
+    this.interfaces = new Interfaces(transpile, interfaces);
+    this.enums = new Enums(transpile, enums);
+  }
+
+  /**
+   * Generate markdown.
+   */
+  public get markdown(): Markdown {
+    const md = new Markdown({ header: { title: "API Reference" } });
```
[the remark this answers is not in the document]
iliapolo: Not sure thats true but i'm open to it. I like having each file define its entire area of responsibility.

src/api/docgen/view/class.ts
```diff
@@ -0,0 +1,72 @@
...
+      header: { title: this.transpiled.name },
+    });
+
+    if (this.klass.interfaces.length > 0) {
+      const ifaces = [];
+      for (const iface of this.klass.interfaces) {
+        const transpiled = this.transpile.type(iface);
+        ifaces.push(`[${Markdown.pre(transpiled.fqn)}](#${transpiled.fqn})`);
+      }
+      md.bullet(`${Markdown.italic("Implements:")} ${ifaces.join(", ")}`);
```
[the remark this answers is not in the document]
iliapolo: perhaps..but perhaps not :)

src/api/docgen/view/documentation.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  constructor(options: DocumentationOptions) {
+    this.assembly = options.assembly;
+
+    switch (options.language) {
+      case "python":
+        this.transpile = new PythonTranspile(this.assembly.system);
+        break;
+      default:
+        throw new Error(
```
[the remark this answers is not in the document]
iliapolo: This is essentially caught by the browser and logged to the console. This does mean that if this happens to a user, the page will simply not render and nothing will happen, if this happens it's a bug - how do we expose this to users? should we? Like I mentioned we need to have a small discussion about error handling in the UI components.

src/api/docgen/view/documentation.ts
```diff
@@ -0,0 +1,114 @@
...
+          `Generating documentation for ${options.language} is not supported`
+        );
+    }
+
+    if (options.submoduleName) {
+      this.submodule = this.findSubmodule(options.submoduleName);
+    }
+
+    this.includeApiReference = options.apiReference ?? true;
+    this.includeReadme = options.readme ?? false;
```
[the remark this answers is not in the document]
iliapolo: Not sure - i'll change it 👍

src/api/docgen/view/documentation.ts
```diff
@@ -0,0 +1,114 @@
...
+  private findSubmodule(submoduleName: string): reflect.Submodule {
+    const submodules = this.assembly.submodules.filter(
+      (s) => s.name === submoduleName
+    );
+
+    if (submodules.length === 0) {
+      throw new Error(`Submodule ${submoduleName} not found`);
+    }
+
+    if (submodules.length > 1) {
```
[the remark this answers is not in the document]
iliapolo: Added. This isn't for submodule selection, but just a way to lookup the submodule object based on the name given by the URL parameter in the UI case.

src/api/docgen/view/parameter.test.ts
iliapolo: 👍

src/setupTests.ts
iliapolo: Yeah i'll change.

src/setupTests.ts
iliapolo: I don't think we need schema validation because these are official schema's produced by JSII, and validation takes a lot of time. 

As far as sanitation and security, yes, but this is out of scope for now I believe..?

.projenrc.js
```diff
@@ -105,3 +115,74 @@ project.eslint.addRules({
 });
 
 project.synth();
+
+function codeGenFetchAssemblies() {
```
eladb, line 119: What's the value of generating this code instead of just writing it directly? Not sure I completely get it

src/Package/index.tsx
```diff
@@ -44,17 +43,85 @@ export function Package() {
...
+async function fetchAssembly(
+  name: string,
+  version: string,
+  scope?: string
+): Promise<spec.Assembly> {
+  if (version.startsWith("^")) {
+    version = version.substring(1, version.length);
+  }
+
+  // e.g https://awscdk.io/packages/@aws-cdk/alexa-ask@1.106.0/jsii.json
```
eladb, line 78: AFAIK we can don't need to fetch from awscdk.io because we have a cloudfront proxy, no?

src/Package/index.tsx
```diff
@@ -44,17 +43,85 @@ export function Package() {
...
+  const response = await fetch(assemblyPath);
+  if (!response.ok) {
+    throw new Error(
+      `Failed fetching assembly for ${assemblyPath}: ${response.statusText}`
+    );
+  }
+  return response.json();
+}
+
+async function fetchAssemblies(
```
eladb, line 89: Perhaps Extract all these functions to a separate file

.projenrc.js
```diff
@@ -80,14 +83,21 @@ const project = new web.ReactTypeScriptProject({
...
 
 // npm tarball will only include the contents of the "build"
 // directory, which is the output of our static website.
-project.npmignore?.addPatterns("!/build");
-project.npmignore?.addPatterns("/public");
+project.npmignore.addPatterns("!/build");
+project.npmignore.addPatterns("/public");
+
+// assemblies used for tests
+project.npmignore.addPatterns("src/__assemblies__");
```
eladb, line 97: These are test fixtures, correct? Maybe just call them `__fixtures__`? 

It's just not clear that these are intended for tests since tests are co located here. 

Alternatively consider putting tests under `src/__tests__` but I am actually quite fond of the co-location

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,193 @@
...
+    let sanitized = line.trim();
+
+    if (line.startsWith("-")) {
+      sanitized = sanitized.substring(1, line.length).trim();
+    }
+
+    return sanitized;
+  }
+
+  public static bold(text: string): string {
```
eladb, line 75: Might be useful to add a custom formatted for underline which is encoded as `_underline_`

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,193 @@
...
+   * Render a `jsii-reflect.Docs` element into the markdown.
+   */
+  public docs(docs: reflect.Docs) {
+    if (docs.summary) {
+      this.lines(Markdown.sanitize(docs.summary));
+      this.lines("");
+    }
+    if (docs.remarks) {
+      this.lines(Markdown.sanitize(docs.remarks));
+      this.lines("");
```
eladb, line 108: I saw in the example you linked that if "remarks" include markdown formatted text (eg bullets), then the output looks odd.

Wondering maybe all remarks should be under a block quote so if they are formatted it will look ok

src/PackageDocs/index.tsx
```diff
@@ -1,15 +1,27 @@
...
+    language: props.language,
+    submoduleName: props.submodule,
+  });
+
+  const source = doc.markdown.render();
+
   return (
     <Box width="100%">
-      <ReactMarkdown skipHtml components={ChakraUIRenderer()} source={readme} />
+      <ReactMarkdown skipHtml components={ChakraUIRenderer()} source={source} />
```
eladb, line 24: Does this include code block syntax highlighting?

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  public section(section: Markdown) {
+    this._sections.push(section);
+  }
+
+  public render(headerSize: number = 0): string {
+    if (headerSize > 6) {
+      // headers are mapped to `h1-h6` html elements.
+      // passed that, markdown jsut renders `#` signs.
```
eladb, line 151:
```suggestion
      // passed that, markdown just renders `#` signs.
```

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,193 @@
...
+      caption = `~~${caption}~~`;
+    }
+
+    if (this.options.header?.sup) {
+      caption = `${caption}<sup>${this.options.header?.sup}</sup>`;
+    }
+
+    return caption;
+  }
+}
```
eladb, line 193: Nice little markdown library

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,323 @@
+import { toSnakeCase } from "codemaker";
+import * as reflect from "jsii-reflect";
+import { Markdown } from "../render/markdown";
+import * as transpile from "./transpile";
+
+/**
+ * A python transpiler.
+ */
+export class PythonTranspile extends transpile.AbstractTranspile {
```
eladb, line 9: We usually use the Base postfix to indicate an abstract base class:

```suggestion
export class PythonTranspile extends transpile.TranspileBase {
```

src/api/docgen/transpile/python.ts
eladb: I believe that's the behavior of the code generator

src/api/docgen/transpile/python.ts
```diff
@@ -0,0 +1,323 @@
...
+    // type is inside this submodule.
+    return submodules[0];
+  }
+
+  private typing(type: "List" | "Mapping" | "Any" | "Union"): string {
+    return `typing.${type}`;
+  }
+
+  private builtins(type: "bool" | "str") {
+    return `builtins.${type}`;
```
eladb, line 263: I understand this is how this is expressed in mypy, but do we really need the `builtins.` prefix everywhere?

Generally speaking: I am not sure that the mypy representation is the most friendly way to present the reference in python.

src/api/docgen/view/__snapshots__/class.test.ts.snap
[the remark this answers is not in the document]
eladb: See https://jestjs.io/docs/snapshot-testing

src/api/docgen/view/api-reference.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.classes = new Classes(transpile, classes);
+    this.structs = new Structs(transpile, interfaces);
+    this.interfaces = new Interfaces(transpile, interfaces);
+    this.enums = new Enums(transpile, enums);
+  }
+
+  /**
+   * Generate markdown.
+   */
+  public get markdown(): Markdown {
```
eladb, line 38: Personally I prefer not to use a getter API for something that preforms processing. I'd call this `render(): Markdown`

src/api/docgen/view/api-reference.ts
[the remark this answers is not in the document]
eladb: Maybe just put in a const at the beginning of this file

src/api/docgen/view/attribute.ts
```diff
@@ -0,0 +1,14 @@
+import * as reflect from "jsii-reflect";
+import { Markdown } from "../render/markdown";
+import { Transpile } from "../transpile/transpile";
+import { Property } from "./property";
+
+export class Attribute {
```
eladb, line 6: Add some docstrings

P.S. what is an attribute? I am not familiar with that term in the JSII world

src/api/docgen/view/class.ts
```diff
@@ -0,0 +1,72 @@
...
+import { Transpile, TranspiledClass } from "../transpile/transpile";
+import { Attributes } from "./attributes";
+import { Constants } from "./constants";
+import { Initializer } from "./initializer";
+import { InstanceMethods } from "./instance-methods";
+import { StaticFunctions } from "./static-functions";
+
+export class Class {
+  public static isConstruct(klass: reflect.ClassType): boolean {
+    if (klass.fqn === "constructs.Construct") return true;
```
eladb, line 12: Const

.projenrc.js
```diff
@@ -105,3 +115,74 @@ project.eslint.addRules({
...
+  const os = require("os");
+  const path = require("path");
+
+  if (packageVersion.startsWith("^")) {
+    packageVersion = packageVersion.substring(1, packageVersion.length);
+  }
+
+  const output = `${packagesPath}/${packageName}@${packageVersion}/jsii.json`;
+
+  const tempDir = fs.mkdtempSync(`${os.tmpdir()}/assembly`);
```
Chriscbr, line 163: Should this temporary directory be cleaned up after the assemblies are copied successfully?

src/api/docgen/render/markdown.ts
```diff
@@ -0,0 +1,193 @@
...
+      // headers are mapped to `h1-h6` html elements.
+      // passed that, markdown jsut renders `#` signs.
+      // lets see if and when we'll hit this limit.
+      throw new Error(`Unable to render markdown. Header limit (6) reached.`);
+    }
+
+    const content: string[] = [];
+    if (this.header) {
+      const heading = `${"#".repeat(headerSize)} ${this.header}`;
+      content.push(`${heading} <a name="${this.id}"></a>`);
```
Chriscbr, line 159: Do empty anchor tags like this have a special semantics in the markdown rendering library you're using? How would these appear to the user vs. making the entire header a link?

.projenrc.js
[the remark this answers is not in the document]
addihorowitz: What do you mean by "dev script"? If it's for dev purpose (and not being used by the production app) then perhaps we should put it on an external "scripts" directory?

src/api/docgen/view/documentation.ts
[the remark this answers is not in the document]
addihorowitz: I wouldn't postpone the merge of this PR for that, but I do think informative error handling should be added - whether by informative response code or by informative exceptions, depending on the scenario

src/setupTests.ts
addihorowitz: You are correct! It's something we will do on the backend

.projenrc.js
```diff
@@ -29,8 +29,11 @@ const project = new web.ReactTypeScriptProject({
     "@emotion/styled@^11",
     "chakra-ui-markdown-renderer",
     "framer-motion@^4",
-    "@uiw/react-markdown-preview",
     "react-router-dom",
+    "@uiw/react-markdown-preview",
```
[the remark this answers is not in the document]
iliapolo: I'm not sure it is the best option. It was the only one I found that gives us a reasonable view without special customizations. It's supposed to be a simple CSS wrapper on top of [`react-markdown`](https://github.com/remarkjs/react-markdown), which is backed by [`remarkjs`](https://remark.js.org/).

But I think going forward we should either stick to pure [`remarkjs`](https://remark.js.org/) libraries or roll our own solution for markdown rendering. I'll make sure we revisit this dependency before launch.

.projenrc.js
iliapolo: Yeah, it started of way more dynamic than what it is now - ejected.

src/Package/index.tsx
iliapolo: Yes, debug leftovers.

src/Package/index.tsx
iliapolo: Moved 👍

.projenrc.js
iliapolo: I'm also quite fond of the co-location. changed to `__fixtures__/assemblies` since we might have other fixtures as well.

src/api/docgen/render/markdown.ts
iliapolo: What do you mean? To my understanding markdown doesn't have underlines. 

`_underline_` => _underline_

src/PackageDocs/index.tsx
iliapolo: Yes

src/api/docgen/render/markdown.ts
iliapolo: I actually started out this way and had some formatting issues as well. Also there are other things in quote right now (`see`, `link`) so it will cause a quote overload. 

Lets iterate on this later on.

src/api/docgen/render/markdown.ts
iliapolo: 👍

src/api/docgen/transpile/python.ts
iliapolo: 👍

src/api/docgen/transpile/python.ts
iliapolo: I don't actually think its related to `mypy` here...i'm not sure why the code generator uses `builtins.str` and not just `str`. I'll ping @RomainMuller about it and in the meantime change to just use `bool` and `str`.

I wonder also if we should remove the type hints from the code snippet, making it much more easy on the eyes, and leaving types only in the dedicated parameter section.

src/api/docgen/render/markdown.ts
eladb: I think it should be possible to add some custom markdown formatting which will render `_this_` with underline in the construct hub. In GitHub, it will render as *italics*.

eladb:
@Chriscbr welcome!

src/api/docgen/transpile/python.ts
eladb: > I wonder also if we should remove the type hints from the code snippet, making it much more easy on the eyes, and leaving types only in the dedicated parameter section.
 
Either omit completely or format in a more human-friendly way

src/api/docgen/view/api-reference.ts
iliapolo: changed to `render`

src/api/docgen/view/class.ts
iliapolo: 👍

src/api/docgen/view/attribute.ts
iliapolo: The current python docs use `Attributes` for instance properties (i.e `instance.some_property`).

In practice there is no different between this and a `reflect.Property` - which is why it simply delegates to the `Property` class. I'm quite font of this distinction because we normally use `Properties` to mean the input properties of a construct, not the output ones.

eladb: But isn't that a Python-specific terminology.

(we use "props" and "properties" to distinguish between initialization props and class properties -- not that I like it but that's the semantic we used so far).

eladb: Also, the term "attribute" is used in the AWS CDK to denote resource attributes (such as ARN), so that might also be conflicting..

eladb: I'd go with `Property` here and if you want a term for the initialization props use `prop`

.projenrc.js
iliapolo: I'm usually not a stickler for it since temp is purged on restart...but ok :)

iliapolo:
@Chriscbr Welcome :)

src/api/docgen/render/markdown.ts
iliapolo: They are a way to override the anchor name that is generated by default. It works both on GitHub and in the library we are using. 

> How would these appear to the user vs. making the entire header a link?

Do you mean doing something like this? 

```ts
content.push(`<a name="${this.id}">${heading} </a>`);
```

If so - this doesn't render properly:

Screen Shot 2021-05-31 at 12 44 00 PM

We also don't want the entire header to be a clickable, we just need to place and anchor next to it so we can link **to** it.

src/api/docgen/render/markdown.ts
eladb: We can add later

src/api/docgen/render/markdown.ts
eladb: Okay. I'd argue that links and see can  also qualify as remarks

eladb (approved):
Ship it. We can iterate on some remaining comments later. Generally structure looks great

src/api/docgen/view/attribute.ts
iliapolo: Ok, i'll just reuse the `Property` class for now and we'll see where that gets us. 

This also brings up the topic of having language specific header titles (i.e `Attributes` in python and `Properties` in TS) which my current design doesn't address. I'll use `Properties` for now since I agree it is probably more generic.

src/api/docgen/render/markdown.ts
Chriscbr: > They are a way to override the anchor name that is generated by default.

I didn't know about that, cool!

.projenrc.js
RomainMuller: I reckon there is a bug in `projen` whereby the `build` task (and I got a similar behavior on `compile`) appears to be reset by some action (was not able to quite pinpoint which though)... And then this causes some task dependencies to... get lost.

.projenrc.js
RomainMuller: Nitpick: I kinda sorta would rather we `npm view` than `npm v`. Not everyone reading this is familiar with all `npm` commands, and abbreviations are mostly meant for when you are typing that in an interactive shell (IMHO).

RomainMuller (approved):

PR #70, opened 2021-06-02
PackageNav

MrArnoldPalmer:
Fixes #

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -136,6 +136,10 @@
       "type": "build"
     },
     {
+      "name": "@chakra-ui/icons",
+      "type": "runtime"
+    },
+    {
       "name": "@chakra-ui/react",
       "type": "runtime"
     },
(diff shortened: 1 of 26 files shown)


src/PackageDocs/index.tsx
```diff
@@ -1,27 +1,140 @@
...
+  };
+
+interface NavItem {
+  title: string;
+  path: string;
+  children: NavItem[];
+}
 export function PackageDocs(props: PackageDocsProps) {
+  const [navItems, setNavItems] = useState<NavItem[]>([]);
   const doc = new Documentation({
```
gabewomble, line 60: I don't know how expensive it is to create an instance of `Documentation`, but assuming that it is, we will probably want to memoize this so that it is not created on every render i.e when `props` change or `navItems` state changes.

I suggest using `useMemo`

```ts
const { assembly, language, submodule } = props; // or destructure in parameters

const doc = useMemo(() => new Documentation({
  assembly,
  language,
  submoduleName: submodule,
}), [assembly, language, submodule]);
```

This will allow us to only need to redefine doc when any of those dependencies are changed

PR #71, opened 2021-06-03
DRAFT - Publisher Area + Operator Area

gabewomble:
Summary

This work begins to flesh out the Publisher & Operator area UI. It is almost complete, but I wanted to begin gathering feedback on the work, as I've implemented a hook - `useRequest`, and overhauled the folder structure.

Main changes:
- Initial Publisher & Operator area with rudimentary load states
- `useRequest` hook, wraps a function which returns a promise (usually for requests), returns the wrapped function and an object representing the request's state
- moved components into `/components` and views into `/views`

Todo:
- I still need to build out the dependency dropdown, and re-format some of the operator area information
- Need to write tests for `useRequest`, and some of the components which render load states

Screen Shot 2021-06-03 at 4 28 13 PM

Loading states - different components use different data, not sure if we should use individual spinners so they don't block each other from displaying, or if we want to use one spinner and wait for all data to be ready

Screen Shot 2021-06-03 at 4 46 16 PM


src/hooks/useRequest.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export type UseRequestReturn<T> = [() => Promise<void>, UseRequestResponse<T>];
+
+/**
+ * Provides an API to statefully interact with promises. This hook
+ * prevents state updates if promise has not resolved before component unmounts.
+ * It will return a tuple with a request() function, and an object representing the state
+ */
+export function useRequest<T>(
```
gabewomble, line 22: This is somewhat a "quick and dirty" implementation, though it is stable. I wanted to implement a standard way of interacting with api's in our client with access to load state & errors sooner than later. Once we have more clarity around our backend, we can review if we want to use a 3rd party dependency such as [react-query](https://react-query.tanstack.com/)

src/utils/url.ts
```diff
@@ -0,0 +1,11 @@
...
+  const params: any = {};
+  const keyValues = search.substring(1, search.length).split("&");
+
+  for (const kv of keyValues) {
+    const parts = kv.split("=");
+    params[parts[0]] = parts[1];
+  }
+
+  return params;
+}
```
MrArnoldPalmer, line 11: I feel like we should just use a library for this personally https://www.npmjs.com/package/qs

src/views/Package/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface PathParams {
+  name: string;
+  scope?: string;
+  version: string;
+}
+
+export function Package() {
+  const { name, scope, version }: PathParams = useParams();
+
+  const [requestMetadata, metadataResponse] = useRequest<Metadata>(() =>
```
MrArnoldPalmer, line 20: I believe TS should be able to use type inference seeing that the return type of `fetchMetadata` is `Promise<MetaData>` and therefore it shouldn't be necessary for us to pass the type param here `useRequest(() => fetchMetaData(name, version. scope))` should work.

src/hooks/useRequest.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const mountedRef = useRef(false);
+
+  useEffect(() => {
+    mountedRef.current = true;
+
+    return () => {
+      mountedRef.current = false;
+    };
+  }, []);
```
MrArnoldPalmer, line 40: I don't understand the point of this function or of `mountedRef` in general. Is this a strategy to delay the execution of the actual request to "next tick" type thing? Feels pretty hacky and why is that necessary?

src/utils/url.ts
gabewomble: We actually can just use the native `URL` web api, this was existing code I had moved

src/views/Package/index.tsx
gabewomble: I've actually re-written this a bit so that the inference is much better

src/hooks/useRequest.ts
gabewomble: The purpose is so that if a promise hasn't resolved when a component is unmounted, we can prevent `setState` calls on unmounted components, which react will throw warnings for

gabewomble: A bit more about this can be found on [this article](https://www.robinwieruch.de/react-warning-cant-call-setstate-on-an-unmounted-component)

MrArnoldPalmer: interesting, ty

src/utils/url.ts
MrArnoldPalmer: gotcha, we don't have to swap it now just a  thought.

gabewomble:
Closing to separate features into own PRs

src/utils/url.ts
gabewomble: @MrArnoldPalmer I'm happy to include these changes in the next PR when my prerequisite changes get merged 🙂

PR #74, opened 2021-06-04
feat: useRequest

gabewomble:
As suggested by @MrArnoldPalmer, I'm breaking out the features of my original PR. This PR only includes the addition of `useRequest` and `@testing-library/react-hooks` so that we can write tests for complex hooks.

The intentions of this hook is to provide the following:
1. A stateful wrapper around promises (primarily intended for API interactions)
2. State update prevention when components unmount while the promise is pending
3. A standard state representation for promises, including loading & error state

Note: A few changes were generated by projen when adding `@testing-library/react-hooks`. The only additions I made were for useRequest


src/hooks/useRequest.ts
```diff
@@ -0,0 +1,90 @@
...
+  (...args: Parameters<T>) => Promise<void>,
+  UseRequestResponse<AwaitedReturn<T>>
+];
+
+/**
+ * Provides an API to statefully interact with promises. This hook
+ * prevents state updates if promise has not resolved before component unmounts.
+ * It will return a tuple with a request() function, and an object representing the state
+ */
+export function useRequest<T extends PromiseFn>(
```
gabewomble, line 30: B/c usage examples will be implemented in the follow-up to this work, I'll provide a quick example below:

```tsx
function UserInfo({ name }: { name: string }) {
  const [requestUserInfo, { data, loading, error }] = useRequest(fetchUserByName);

  useEffect(() => {
    void requestUserInfo(name);
  }, [name]);

  if (loading) return <Loader />;

  if (error) return <Error error={error} />;

  return <UserDetails user={data} />;
}
```

MrArnoldPalmer (approved):

PR #75, opened 2021-06-04
feat: cypress test runner

gabewomble:
This PR introduces the Cypress test runner to our application! It only has a basic configuration at this time. We will want to setup a github action to run these tests on PRs and merges to master, but the effort required warrants a separate PR.

Screen Shot 2021-06-04 at 3 04 53 PM


src/Header/index.tsx
```diff
@@ -2,7 +2,15 @@ import { Box, Flex, Heading, Input, Text } from "@chakra-ui/react";
...
-    <Box borderRadius="md" w="100%" px={1} py={3} color="white" bg="purple.600">
+    <Box
+      bg="purple.600"
+      borderRadius="md"
+      color="white"
+      data-test="header"
+      px={1}
+      py={3}
+      w="100%"
+    >
```
MrArnoldPalmer, line 13: Seems like this change is unrelated.

MrArnoldPalmer (changes requested):
just removal of the change in the Header component then 👍🏻

src/Header/index.tsx
gabewomble: Only addition is the `data-test` attr which made the Header exceed our character limit for the line

MrArnoldPalmer (approved):

PR #76, opened 2021-06-04
feat: storybook addons

gabewomble:
This PR introduces a couple new Storybook addons and fixes the theming in storybook ([solution here](https://github.com/chakra-ui/chakra-ui/issues/2527#issuecomment-728161743))

Addons (screenshots are from our storybook):

[a11y](https://storybook.js.org/addons/@storybook/addon-a11y)

Screen Shot 2021-06-04 at 3 22 28 PM

[storysource](https://storybook.js.org/addons/@storybook/addon-storysource)

Screen Shot 2021-06-04 at 3 23 28 PM

[performance](https://storybook.js.org/addons/storybook-addon-performance)

Screen Shot 2021-06-04 at 3 25 19 PM


.storybook/main.js
```diff
@@ -1,11 +1,30 @@
...
+  ],
+  webpackFinal: async (config) => {
+    return {
+      ...config,
+      resolve: {
+        ...config.resolve,
+        alias: {
+          ...config.resolve.alias,
+          "@emotion/core": toPath("node_modules/@emotion/react"),
+          "emotion-theming": toPath("node_modules/@emotion/react")
```
MrArnoldPalmer, line 25: This is really funky, but I guess makes sense.

.storybook/preview.js
```diff
@@ -1,3 +1,16 @@
+import { withPerformance } from "storybook-addon-performance";
+import { Theme } from "../src/Theme";
+import { CSSReset } from "@chakra-ui/react";
+
+export const decorators = [
+  withPerformance,
+  (Story) => (
+    <Theme>
+      <CSSReset />
```
MrArnoldPalmer, line 9: We don't currently use the reset in the app right? We probably should.

MrArnoldPalmer (approved):

.storybook/preview.js
gabewomble: Good callout - I'll bring that in too

.storybook/main.js
gabewomble: Yeah, it's weird. I linked it in the description above, but found this solution from the [issue here](https://github.com/chakra-ui/chakra-ui/issues/2527#issuecomment-728161743).

PR #77, opened 2021-06-05
feat: package page navigation

### LINKED ISSUE #44: Implement: Package page: package data area - navigation area
Create a Markdown based navigation component to be used in the package page. 

The main content of the package page is a markdown that includes both the user defined `README.md`, as well as a generated API reference doc based on the jsii assembly file. 

Navigation between the different markdown headers should be done via a separate react component that renders indented header links. 

Just for inspiration, GitHub offers this functionality by default for every page that contains a markdown file:

> https://github.com/aws/aws-cdk/tree/master/packages/%40aws-cdk/aws-eks

Screen Shot 2021-05-23 at 8 55 54 PM

## User Actions

* When the user *clicks an item in the Navigation Area*, the *Content Area* displays the relevant paragraph for this item.

## Notes

- Ideally, the component should be purely representational, without any domain specific logic. 

## References

- [markdown-navbar](https://www.npmjs.com/package/markdown-navbar) is a react component that does just that and is worth exploring. From a quick look, and might not be flexible enough for us. 
- Nice example: https://registry.terraform.io/providers/hashicorp/aws/latest/docs

MrArnoldPalmer:
Defines basic components for package page left nav and logic for
extracting link titles and paths from the body rendered from markdown.

Changes the output of the generated markdown in the API reference to
allow passing all relevant header data as html data attributes. This
allows us to reliably query for header content and link paths and
extract the relevant data.

Sanitized the IDs of header links so they are valid. An `FQN` on its own
cannot be used as an anchor link. This logic will need to be standard so
that we can reliably compute links to types in other modules.

Replaces usage of `@uiw/react-markdown-preview` because this broke
header link rendering. It would automatically render an <a /> tag with
an ID, but this ID was generated from the content within the header,
which in some cases included html, code blocks, etc. Rendering our own
links via passing a custom header component resulted in 2 <a /> tags
being rendered, the malformed one and the correct one. This happened
with or without the usage of the `ChakraUIRenderer`.

Fixes #44

PackageNav

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,17 +64,18 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@chakra-ui/icons": "*",
     "@chakra-ui/react": "*",
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
     "@jsii/spec": "^1.30.0",
-    "@uiw/react-markdown-preview": "^3.0.6",
     "chakra-ui-markdown-renderer": "^3.0.1",
     "codemaker": "^1.30.0",
     "framer-motion": "^4",
     "jsii-reflect": "^1.30.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "react-markdown": "^6.0.2",
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.0",
     "web-vitals": "^1.1.2"
--- a/src/components/PackageDocs/Body.tsx
+++ b/src/components/PackageDocs/Body.tsx
@@ -1,0 +1,68 @@
+import { LinkIcon } from "@chakra-ui/icons";
+import { Heading, As } from "@chakra-ui/react";
+import ChakraUIRenderer from "chakra-ui-markdown-renderer";
+import React from "react";
+import ReactMarkdown from "react-markdown";
+import { sanitize } from "../../util/sanitize-anchor";
+
+type HeadingResolverProps = {
+  level: number;
+  children: React.ReactNode;
+};
+
+function Headings({ level, children }: HeadingResolverProps) {
+  const size: string = ["2xl", "xl", "lg", "md", "sm", "xs"][level - 1];
+  const elem = `h${level}` as As<any>;
+
+  // Use DOMParser to look for data attribute for link ID
+  const parser = new DOMParser();
+  const doc = parser.parseFromString(
+    React.Children.toArray(children).join(""),
+    "text/html"
+  );
+
+  const dataElement = doc.querySelector(
+    "span[data-heading-title][data-heading-id]"
+  ) as HTMLElement;
+  const title =
+    dataElement?.dataset.headingTitle ??
+    React.Children.toArray(children)
+      .reduce((accum: string, child): string => {
+        if (typeof child === "string") {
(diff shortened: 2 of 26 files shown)


src/components/PackageDocs/Body.tsx
```diff
@@ -0,0 +1,68 @@
...
+    dataElement?.dataset.headingTitle ??
+    React.Children.toArray(children)
+      .reduce((accum: string, child): string => {
+        if (typeof child === "string") {
+          return `${accum}${child}`;
+        }
+        return accum;
+      }, "")
+      .trim();
+  const id = dataElement?.dataset.headingId ?? sanitize(title);
```
iliapolo, line 37: Are you sure this is working properly? I'm still seeing titles as anchor ids, not `FQN`s:

Screen Shot 2021-06-07 at 12 34 03 PM

Also - are the defaults being applied here to handle the user defined README? Would be good to document them and ideally merge to a single branch.

src/components/PackageDocs/Body.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <Heading my={4} level={level} as={elem} size={size}>
+      {level < 4 && (
```
iliapolo, line 41: Why the restriction? I'm pretty sure we want to show all levels in the navigation...

src/components/PackageNav/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+        );
+      })}
+    </Flex>
+  );
+}
+
+export interface PackageNavItem {
+  title: string;
+  path: string;
+  level: number;
```
iliapolo, line 32: Is this actually used somewhere in this component?

package.json
```diff
@@ -64,17 +64,18 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@chakra-ui/icons": "*",
```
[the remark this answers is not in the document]
MrArnoldPalmer: I'm not sure, but we can look into it separately.

src/components/PackageDocs/Body.tsx
MrArnoldPalmer: Fixed this, it wasn't parsing the `span` elements generated from the markdown render correctly. 😄

src/components/PackageDocs/Body.tsx
MrArnoldPalmer: We can, it just feels extremely crowded and each construct has "initializer + methods + properties" and so there are a lot of duplicate titles in the left nav. I'd say perhaps once the menu is dynamic, IE collapses/expands, it may make it feels less crowded. But since there are like 6 levels it feels like that will also be hard to keep track of for the user.

src/components/PackageNav/index.tsx
MrArnoldPalmer: The `level` property? No, it just greatly simplifies the recursive building of the nav items. Additionally I do believe it's likely to be used in the future as the level will dictate the styling of the elements.

MrArnoldPalmer:
damn, when force pushing a rebase I accidentally deleted the branch. Reopened https://github.com/cdklabs/construct-hub-webapp/pull/78

PR #78, opened 2021-06-07
feat: package page navigation

### LINKED ISSUE #44: Implement: Package page: package data area - navigation area

MrArnoldPalmer:
Defines basic components for package page left nav and logic for
extracting link titles and paths from the body rendered from markdown.

Changes the output of the generated markdown in the API reference to
allow passing all relevant header data as html data attributes. This
allows us to reliably query for header content and link paths and
extract the relevant data.

Sanitized the IDs of header links so they are valid. An `FQN` on its own
cannot be used as an anchor link. This logic will need to be standard so
that we can reliably compute links to types in other modules.

Replaces usage of `@uiw/react-markdown-preview` because this broke
header link rendering. It would automatically render an <a /> tag with
an ID, but this ID was generated from the content within the header,
which in some cases included html, code blocks, etc. Rendering our own
links via passing a custom header component resulted in 2 <a /> tags
being rendered, the malformed one and the correct one. This happened
with or without the usage of the `ChakraUIRenderer`.

Fixes #44

PackageNav


iliapolo (approved):

src/components/PackageDocs/Body.tsx
```diff
@@ -0,0 +1,80 @@
...
+    React.Children.toArray(children)
+      .reduce((accum: string, child): string => {
+        if (typeof child === "string") {
+          return `${accum}${child}`;
+        }
+        return accum;
+      }, "")
+      .trim();
+  /* console.log(dataElement?.dataset); */
+  const id = dataElement?.dataset.headingId ?? sanitize(title);
```
gabewomble, line 43: For follow up revision: can we separate these declarations for readability & omit the console.log?

Also, it looks like we'll want to optionally access `dataElement?.dataset` properties since we're assuming that the dataElement can be undefined

PR #80, opened 2021-06-07
feat: React eslint rules

gabewomble:
Configures a few baseline eslint plugins for react so that we can enforce best practices in our development lifecycle.

Additions:
`eslint-plugin-react`
`eslint-plugin-react-hooks`
`eslint-plugin-jsx-a11y`

We did not have any a11y violations so here is a screenshot showing what the a11y rules do!
Screen Shot 2021-06-07 at 2 17 29 PM


src/views/Home/index.tsx
```diff
@@ -4,15 +4,15 @@ import "./style.css";
...
+      <img alt="logo" className="App-logo" src={logo} />
       <p>
         Edit <code>src/App.tsx</code> and save to reload.
       </p>
       <a
         className="App-link"
         href="https://reactjs.org"
-        target="_blank"
         rel="noopener noreferrer"
+        target="_blank"
```
MrArnoldPalmer, line 15: 👍🏻

MrArnoldPalmer (approved):

Issue #45, opened 2021-05-23
Implement: Package page - language bar


iliapolo:
The Getting Started Area is part of the Package Page Header. It displays information for users that may not have significant information if at all with Constructs and allows them to set their preferred programming language. This should be mostly useful for users who fount the package page through an external search engine (like Google).

## User Actions

* Each construct library/package supports at least one programming language. A user can set their preferred programming language out of the *supported (enabled) programming languages*. The programming languages that are not supported by the displayed package will be *disabled for clicking.*
    * The README, API reference and code samples will be rendered accordingly. 
    * In addition, a *local* *Cookie* will be updated with the programming language that was selected. 
* Make sure to ask for cookie consent before accessing the cookie. If no language is selected, the default is typescript

iliapolo:
> A user can click development quickstart. A new tab will be open with with the relevant CDK quickstart guide. For example, if the construct is of AWS-CDK, it will open the quickstart of AWS-CDK. If it’s a construct of CDK8s, it will open the quickstart guide for CDK8s.

We dont have this information right now in a structured way. jsii does not enforce nor extract any specific key from the `package.json` that may indicate this. Decide what to do.

gabewomble:
Just to clarify a few behaviors in this issue:

- We will be storing a value in localStorage to persist a user's preferred language (not in a cookie)

- The order of priority for a selected language is the following
1. If a valid language is found in a query param in the URL use that first. This will not persist / overwrite a user's preferred language
2. If a valid language is found in localStorage, use that language
3. Otherwise fallback to a default language - which is typescript for now.

We will only update the user preferred language when a user interacts with the language bar.

Something to note is that if a construct only supports a specific language, or the supported languages do not include a user's preferred language, we select the first language the construct supports without overwriting user preferences

PR #85, opened 2021-06-08
feat: language bar component

gabewomble:
This work only implements the LanguageBar component, but does not consume it in the app. To test the component, you can launch storybook with `yarn storybook`. This is a prerequisite to #45 

Screen Shot 2021-06-08 at 1 34 13 PM
Screen Shot 2021-06-08 at 1 34 19 PM
Screen Shot 2021-06-08 at 1 39 21 PM

Disabled languages hidden:
Screen Shot 2021-06-08 at 1 48 04 PM

A11y Tests:
Screen Shot 2021-06-08 at 1 48 46 PM


cypress/support/commands.js
```diff
@@ -24,4 +24,4 @@
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite('visit', (originalFn, url, options) => { ... })
 
-Cypress.Commands.add('getByDataTest', (dataTest) => cy.get(`[data-test="${dataTest}"]`));
+Cypress.Commands.add('getByDataTest', (dataTest) => cy.get(`[data-testid="${dataTest}"]`));
```
gabewomble, line 27: this is just so that we use the same test id attributes across unit tests and cypress tests

MrArnoldPalmer (approved):
I like the ability to show all languages even if not enabled and make them not selectable. Overall looks good we can implement on the page and show to the team before deciding whether design changes are needed. Code and tests LGTM for a first pass though.

PR #86, opened 2021-06-08
bug: fix language bar icons

gabewomble:
Silly mistake happened when combining separate `NAME_MAP` and `ICON_MAP` objects into a single `LANGUAGE_MAP`. Didn't realize the icons were pointing at the wrong assets 😅


MrArnoldPalmer (approved):
whoops, neither did I lol

PR #90, opened 2021-06-10
feat: search

### LINKED ISSUE #5: Implement: Search
The search feature is implemented client-side, where the `catalog.json` object is loaded on-demand, then filtered in-memory for search purposes.

The results are shown in a paginated view.

MrArnoldPalmer:
Implements initial logic for searching packages from the catalog in the
awscdk.io `packages.json` file. Uses the same filter logic as
awscdk.io's current search.

Stores search field input in a query param to make results linkable.
Currently requires a form submit to update results since this
potentially navigates users away from their current page and shouldn't
be triggered just input.

Abstract's reusable logic for parsing query string to the `useQuery`
hook and removes the custom parser in favor of `URLSearchParams`.

Fixes #5

### DIFF
--- a/src/api/package/packages.ts
+++ b/src/api/package/packages.ts
@@ -1,0 +1,43 @@
+export interface Packages {
+  packages: {
+    name: string;
+    version: string;
+    metadata: {
+      name: string;
+      description: string;
+      scope: string;
+      version: string;
+      keywords: string[];
+      date: string;
+      links: {
+        npm: string;
+        homepage: string;
+        repository: string;
+        bugs: string;
+      };
+      author: {
+        name: string;
+        url: string;
+      };
+      publisher: {
+        username: string;
+        email: string;
+      };
+      mantainers: {
+        username: string;
+        email: string;
+      }[];
+    };
+  }[];
+}
+
+export async function fetchPackages(): Promise<Packages> {
+  const response = await fetch("/index/packages.json");
+
+  if (!response.ok) {
+    console.error(response.statusText);
+    throw new Error(`Failed fetching packages index: ${response.statusText}`);
+  }
+
+  return response.json();
+}
--- a/src/api/package/packages.ts
+++ b/src/api/package/packages.ts
@@ -1,33 +1,10 @@
+import { Metadata } from "./metadata";
+
 export interface Packages {
   packages: {
     name: string;
     version: string;
-    metadata: {
-      name: string;
-      description: string;
-      scope: string;
-      version: string;
-      keywords: string[];
-      date: string;
-      links: {
-        npm: string;
-        homepage: string;
-        repository: string;
-        bugs: string;
-      };
-      author: {
-        name: string;
-        url: string;
-      };
-      publisher: {
-        username: string;
-        email: string;
-      };
-      mantainers: {
-        username: string;
-        email: string;
-      }[];
-    };
+    metadata: Metadata;
   }[];
 }
 
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -48,7 +48,6 @@ export function Header() {
               placeholder="search"
               value={searchValue}
             />
-            {searchValue}
           </form>
         </Box>
         <Flex>
--- a/src/components/Header/index.tsx
+++ b/src/components/Header/index.tsx
@@ -1,7 +1,20 @@
 import { Box, Flex, Heading, Input, Text } from "@chakra-ui/react";
+import { useState } from "react";
+import { useHistory } from "react-router-dom";
+import { useQuery } from "../../hooks/useQuery";
 import { Logo } from "../../icons/Logo";
 
 export function Header() {
+  const q = useQuery().get("q") ?? "";
+  const [searchValue, setSearchValue] = useState(q);
+  const history = useHistory();
+  const onSearchChange = (evt: React.ChangeEvent<HTMLInputElement>): void => {
+    setSearchValue(evt.target.value);
+  };
+  const onSearchSubmit = (evt: React.FormEvent): void => {
+    evt.preventDefault();
+    history.push(`/search?q=${searchValue}`);
+  };
   return (
     <Box
       bg="white"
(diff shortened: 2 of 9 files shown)


src/api/package/packages.ts
```diff
@@ -0,0 +1,43 @@
...
+        url: string;
+      };
+      publisher: {
+        username: string;
+        email: string;
+      };
+      mantainers: {
+        username: string;
+        email: string;
+      }[];
```
gabewomble, line 29: I believe this can actually just use the `Metadata` interface from  the adjacent `metadata.ts` file

src/hooks/useQuery.ts
```diff
@@ -0,0 +1,5 @@
+import { useLocation } from "react-router-dom";
+
+export function useQuery() {
+  return new URLSearchParams(useLocation().search);
+}
```
gabewomble, line 5: it may be a good idea to rename this to `useQueryParams` as this name is very similar to one of the main hooks implemented by [react-query](https://react-query.tanstack.com/guides/queries). This change would be primarily to prevent confusion from contributors

src/views/SearchResults/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }, [requestPackages]);
+
+  const results = useMemo(() => {
+    /* const searcher = new FuzzySearch(packagesResponse.data?.packages ?? [], [ */
+    /*   "name", */
+    /* ]); */
+
+    /* return query ? searcher.search(query) : []; */
+    return data?.packages.filter((item) =>
+      JSON.stringify(item).includes(query)
```
gabewomble, line 23: I wonder if it makes sense to revise how we implement this search. I think at a basic level we could just filter by `name.toLowerCase().includes(query.toLowerCase())` which would be less expensive than JSON.stringifying every item. We could even expand on this by doing something similar to github's filters with other assembly properties such as `keywords` or `author`.

A few ways we could do this - 
a: github style: `keyword:eks author:foo @foo/eks-cdk-package`

b: Search bar only for names, add tags for key - value filters (don't have a good ex off the top of my head, though I've seen it in a few places before)

src/views/SearchResults/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  useEffect(() => {
+    void requestPackages();
+  }, [requestPackages]);
+
+  const results = useMemo(() => {
+    /* const searcher = new FuzzySearch(packagesResponse.data?.packages ?? [], [ */
+    /*   "name", */
+    /* ]); */
+
+    /* return query ? searcher.search(query) : []; */
```
gabewomble, line 21: ✂️

gabewomble (changes requested):
Overall looks great! Left a few comments & looking forward to testing changes in the morning

src/views/SearchResults/index.tsx
MrArnoldPalmer: I actually used the `fuzzy-search` library at first but ended up doing this as it's what `awscdk.io` currently uses. I agree this isn't a long term solution but for now I think this is mostly about setting up the lifecycle of input -> submit -> show results page.

gabewomble: & one more thought here - if the query is an empty string we can just return `data?.packages ?? []` as a slight optimization

src/components/Header/index.tsx
```diff
@@ -1,7 +1,20 @@
...
+  const q = useQuery().get("q") ?? "";
+  const [searchValue, setSearchValue] = useState(q);
+  const history = useHistory();
+  const onSearchChange = (evt: React.ChangeEvent<HTMLInputElement>): void => {
+    setSearchValue(evt.target.value);
+  };
+  const onSearchSubmit = (evt: React.FormEvent): void => {
+    evt.preventDefault();
+    history.push(`/search?q=${searchValue}`);
+  };
```
gabewomble, line 17: We may eventually want to break out the search logic & ui into a separate `<Search />` component

MrArnoldPalmer: yeah I agree, lets do it with the UI implementation?

src/contexts/Search.tsx
```diff
@@ -0,0 +1,3 @@
+import React from "react";
+
```
gabewomble, line 2: looks like this context isn't being used

src/views/SearchResults/index.tsx
MrArnoldPalmer: so the current implementation is if the query is empty, show all results. I say we stick with that for now.

gabewomble: I agree with that - I just mean that we can skip the filter + JSON.stringify on the package list since we know we'll just return everything!

MrArnoldPalmer: ahhh gotcha, yeah sounds good!

src/views/SearchResults/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    );
+  }, [query, data]);
+
+  return (
+    <SimpleGrid columns={[1, null, 3, null, 5]} p={6} spacing={6}>
+      {results &&
+        results.map((pkg) => {
+          return (
+            <Link key={pkg.name} to={`/packages/${pkg.name}/v/${pkg.version}`}>
+              <GridItem
```
gabewomble, line 28: Another idea for another time - we'll probably want to split this markup into two separate components: `<PackageList />` and `<PackagePreview />`, that way we can storybook & test them separately. I imagine the component api would be as simple as `<PackageList packages={results} />

src/components/Header/index.tsx
```diff
@@ -27,7 +40,16 @@ export function Header() {
...
-          <Input id="search" name="search" placeholder="search" />
+          <form onSubmit={onSearchSubmit}>
+            <Input
+              id="search"
+              name="search"
+              onChange={onSearchChange}
+              placeholder="search"
+              value={searchValue}
+            />
+            {searchValue}
```
gabewomble, line 51: just a nitpick but it seems a little odd mirroring the searchValue right under the search bar - could we remove this?

MrArnoldPalmer: Oh whoops yes this was for debugging purposes. Out it comes!

gabewomble (approved):

PR #96, opened 2021-06-14
bug: fix safari issue caused by codemaker

gabewomble:
`codemaker`'s dependency on decamelize was causing the page to error out on safari. Since `codemaker` was only used to `snake_case` some text, I replaced the dependency with a browser friendly case library. The web app should have no issues running on safari now 🙂


package.json
```diff
@@ -66,7 +66,7 @@
     "json-schema": "^0.3.0",
     "npm-check-updates": "^11",
     "prettier": "^2.3.1",
-    "projen": "^0.24.6",
+    "projen": "^0.22.15",
```
RomainMuller, line 69: I suppose this rollback was not intentional...

RomainMuller (changes requested):

package.json
MrArnoldPalmer: make sure to delete `yarn.lock` and `node_modules` then remove changes from package.json, then run `yarn install --frozen-lockfile` before doing `yarn projen`. If you have an old version of projen locally that you're running against it will downgrade.

MrArnoldPalmer (approved):

PR #97, opened 2021-06-14
feat: Language Bar UI

gabewomble:
Additional progress for #45 
This work sets up the Getting Started UI on the package profile page, and implements language selection persistence via localStorage. The rules for selected language are as follows:

1. Use language from query param in URL if present (will not be stored)
2. Fall back to language from localStorage, or use the default "ts"
3. When a user manually changes language, persist the selection in localStorage

I've continued shifting the layout & UI to a card-based design on a simple grid:

Screen Shot 2021-06-14 at 3 20 36 PM
Screen Shot 2021-06-14 at 3 21 18 PM


src/components/GettingStarted/DisabledLangPopover.tsx
```diff
@@ -0,0 +1,34 @@
...
+          icon={<QuestionOutlineIcon />}
+          variant="ghost"
+        />
+      </PopoverTrigger>
+      <PopoverContent>
+        <PopoverArrow />
+        <PopoverCloseButton />
+        <PopoverHeader>Why are some languages disabled?</PopoverHeader>
+        <PopoverBody>
+          This construct has not yet implemented libraries for all available
```
gabewomble, line 28: Looking for suggestions on how to better phrase this. I wanted to provide a disclaimer for users to explain why some languages may be disabled so that they would not be confused

src/components/LanguageBar/index.tsx
```diff
@@ -58,50 +48,52 @@ export function LanguageBar({
   showDisabled = false,
 }: LanguageBarProps) {
   return (
-    <Flex data-testid="language-bar" justify="center">
-      {(showDisabled ? ALL_LANGUAGES : targetLanguages).map((language) => {
-        const isDisabled = !targetLanguages.includes(language);
-        const isSelected = language === selectedLanguage;
+    <Flex data-testid="language-bar">
+      {(showDisabled ? LANGUAGES : targetLanguages).map(
+        (language: Language) => {
```
gabewomble, line 53: Only change here is indentation as I added a type annotation

src/components/PublisherArea/index.tsx
```diff
@@ -1,20 +0,0 @@
-import { Flex } from "@chakra-ui/react";
-import type { Metadata } from "../../api/package/metadata";
-import { PackageHeader } from "../PackageHeader";
-
-export interface PublisherAreaProps {
-  metadata: Metadata;
-}
-
-export function PublisherArea({ metadata }: PublisherAreaProps) {
```
gabewomble, line 9: This wrapping component was useless

src/components/PackageDocs/index.tsx
```diff
@@ -42,6 +43,10 @@ export function appendItem(
   }
 }
 
+// We want the nav to be sticky, but it should account for the sticky heading as well
+// The calculation here is heading height (72px) + margin-top (16px).
+const TOP_OFFSET = "88px";
```
MrArnoldPalmer, line 48: I think this should either be a constant that is used in both components to set the header height and the navbar offset, or they should both just use common theme values ie `height={4}` & `top={4}`.

I'm not sure if we even want the header to be globally sticky have we really defined that?

src/contexts/Language/Language.tsx
```diff
@@ -0,0 +1,72 @@
...
+const isValidLang = (lang: Language | null) => lang && LANGUAGES.includes(lang);
+
+export const LanguageContext = createContext<
+  [Language, (language: Language) => void]
+>([defaultLang, () => {}]);
+
+export const useLanguage = () => useContext(LanguageContext);
+
+export function LanguageProvider({ children }: { children: ReactChild }) {
+  const langFromParams = useQueryParams().get(PARAM_KEY) as Language | null;
```
MrArnoldPalmer, line 26:
```suggestion
  const langFromParams = useQueryParams().get(PARAM_KEY) as Language | void;
```

We tend to prefer `undefined` over null and not mixing between them.

iliapolo (approved):

MrArnoldPalmer (approved):
None of these are blocking. I also want to explore if the language context definition can be simplified. I feel like this could just be a hook without a context and state since its sourced from the url + localstorage.

useLanguage() => [language, setLanguage] where set language, just changes the query param and updates localStore, and language just reads from sources in priority order.

src/contexts/Language/Language.tsx
gabewomble: Makes sense. The annotation from the hook was `string | null` so really just wanted to cast the string portion. Wil update!

src/components/PackageDocs/index.tsx
gabewomble: I believe it is a better experience to keep a sticky header. Could def go about setting a theme value for the header height so that this interaction uses the theme

src/components/DependencyDropdown/index.tsx
```diff
@@ -18,9 +19,19 @@ export interface DependencyDropownProps {
...
 
-function DropdownButton(props: PropsOf<typeof Button>) {
-  return <Button {...props} bg="white" size="md" variant="outline" />;
-}
+const DropdownButton = forwardRef((props: PropsOf<typeof Button>, ref) => {
+  return (
+    <Button
+      {...props}
+      bg="white"
+      ref={ref as any}
```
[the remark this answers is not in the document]
gabewomble: turns out I just needed to use the `forwardRef` from chakra UI in this case

PR #99, opened 2021-06-15
fix: type of arguments in code snippets use fqn instead of name

iliapolo:
Currently, the types of arguments we display in the code snippets use the fqn:

```python
def add_masters_role(role: aws_cdk.aws_iam.IRole, 
                     username: str = None)
```

While this is technically correct, it has two problems:

1. It's a little overwhelming, especially since the full type information is also shown in the arguments breakdown just below.
2. In TypeScript, this will be illegal code:

```typescript
def addMastersRole(role: @aws_cdk/aws-iam.IRole, username: string)
```

This PR changes it so that type information shown in the code snippets always use the simple name, which will work in both languages.

```python
def add_masters_role(role: IRole, 
                     username: str = None)
```

```typescript
def addMastersRole(role: IRole, username: string)
```

This change required a much needed refactoring of the way type references are currently transpiled. Instead of rendering the string representation alongside the type information extraction, we split those into two different invocations.

Type information is being collected the same way, regardless of how its going to be displayed. Rendering the string representation of the type reference is now the responsibility of the view components, and is facilitated by a `typeReference.toString()` method that accepts various formatters.


src/api/docgen/transpile/python.ts
```diff
@@ -258,14 +225,15 @@ export class PythonTranspile extends transpile.TranspileBase {
...
-  transpiled: transpile.TranspiledParameter | transpile.TranspiledProperty
-): string {
-  return `${transpiled.name}: ${transpiled.typeReference.raw}${
-    transpiled.optional ? " = None" : ""
-  }`;
+  private formatInput(
+    transpiled: transpile.TranspiledParameter | transpile.TranspiledProperty
+  ): string {
+    const tf = transpiled.typeReference.toString({
+      typeFormatter: (t) => t.name,
```
iliapolo, line 233: The ability to do this is essentially the crux of this PR

eladb (approved):

PR #101, opened 2021-06-15
feat: language bar feedback

gabewomble:
Implements feedback from previous PR


src/hooks/useLanguage/useLanguage.ts
```diff
@@ -0,0 +1,80 @@
...
+export interface UseLanguageOptions {
+  /**
+   * Syncs the preferred language to a query param in URL
+   */
+  updateUrl?: boolean;
+  /**
+   * Saves the selected language to localStorage on select
+   */
+  updateSaved?: boolean;
+}
```
gabewomble, line 40: This was changed to a hook rather than context per @MrArnoldPalmer's suggestion as all the info it needs regarding initial state exists within the URL or localStorage.

The usecases for these options are the following:

updateUrl - package page will always have lang in URL (meaning sharing links will provide a consistent experience)

updateSaved - eventually we will want to create a user preference page where a user can set a preferred theme. As such, we'll no longer want to overwrite that on the package page

MrArnoldPalmer (approved):
❤️ love it!

PR #105, opened 2021-06-16
fix: package-doc rerender performance

gabewomble:
Updates to the url language were re-rendering the page markdown even when the markdown source was unchanged, which was causing poor performance - especially on larger markdown files. This change memoizes the markdown generation, markdown rendering, and the `useQueryParams` hook to minimize unnecessary re-renders


iliapolo:
@gabewomble 

> Updates to the url language were re-rendering the page markdown even when the markdown source was unchanged

I'm not sure I follow. When the language changes the readme source **does** change.

MrArnoldPalmer (approved):

PR #106, opened 2021-06-16
feat: supported languages only

gabewomble:
This work prevents languages that have not yet been implemented from being selected via language bar or query param to prevent doc generation from erroring. This also updates `useLanguage` to always subscribe to the language query param if it is present

### DIFF
--- a/src/hooks/useLanguage/useLanguage.test.ts
+++ b/src/hooks/useLanguage/useLanguage.test.ts
@@ -1,5 +1,6 @@
 import { renderHook } from "@testing-library/react-hooks";
 import { useLocation as useLocationMock } from "react-router-dom";
+import * as languageConstants from "../../constants/languages";
 import { useLanguage } from "./useLanguage";
 
 jest.mock("react-router-dom", () => ({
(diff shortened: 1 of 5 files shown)


src/hooks/useLanguage/useLanguage.test.ts
```diff
@@ -22,6 +23,8 @@ describe("useLanguage", () => {
   const getItem = jest.spyOn(window.localStorage.__proto__, "getItem");
 
   beforeEach(() => {
+    // @ts-ignore
+    languageConstants.TEMP_SUPPORTED_LANGUAGES = languageConstants.LANGUAGES;
```
gabewomble, line 27: For the sake of the test, it makes sense to replace TEMP_SUPPORTED_LANGUAGES with all languages. Otherwise none of the test cases will be valuable since the only supported language is `python`

src/hooks/useLanguage/useLanguage.ts
```diff
@@ -51,7 +57,23 @@ export function useLanguage(options: UseLanguageOptions = {}) {
...
+
+  // Syncs language changes to URL if updateUrl = true
+  useEffect(() => {
+    if (langFromParams !== language && updateUrl) {
+      params.set(PARAM_KEY, language);
+      replace({ pathname, hash, search: params.toString() });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [language, updateUrl]);
+
```
MrArnoldPalmer, line 76: 👍🏻

MrArnoldPalmer (approved):

PR #130, opened 2021-06-23
chore: remove inline runtime chunk

MrArnoldPalmer:
Removes inline script generation using `INLINE_RUNTIME_CHUNK=false`
configuration value in .env file. Adds `.env.local` file to gitignore to
allow local overriding of environment variables for development
purposes.


.projenrc.js
```diff
@@ -195,3 +196,15 @@ function rewireCRA(craTask) {
     }
   }
 }
+
+/**
+ * Add build time configuration values for react-scripts.
+ * Use an `.env.local` file to override for local development.
+ */
+function addBuildConfig() {
```
eladb, line 204: Would be nice to add to projen

eladb: https://github.com/projen/projen/issues/906

eladb (approved):

PR #131, opened 2021-06-23
chore: log doc generation timing

iliapolo:
I've grown weary from trying to guess exactly how much time doc generation takes with the profiler. 

Added a log to make simple.

### DIFF
--- a/src/components/PackageDocs/index.tsx
+++ b/src/components/PackageDocs/index.tsx
@@ -60,7 +60,10 @@ export function PackageDocs({
   const hasApiReference = !isDev || (isDev && q.get("apiRef") !== "false");
 
   const source = useMemo(() => {
-    const startTime = Date.now();
+    const timeLabel = `Timer | docgen(${assembly.name}${
+      submodule ? `.${submodule}` : ""
+    })`;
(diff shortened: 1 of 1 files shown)


src/components/PackageDocs/index.tsx
```diff
@@ -68,7 +69,9 @@ export function PackageDocs({
     });
 
     const md = doc.render();
-    return md.render();
+    const s = md.render();
+    console.log(`Documentation generated in ${Date.now() - startTime}ms`);
```
eladb, line 73: You can use `console.time()`

iliapolo: Cool. Done.

src/components/PackageDocs/index.tsx
```diff
@@ -68,7 +72,9 @@ export function PackageDocs({
     });
 
     const md = doc.render();
-    return md.render();
+    const s = md.render();
+    console.timeEnd(timeLabel);
```
iliapolo, line 76: Example output: `Timer | docgen(@aws-cdk/aws-ecr): 14.9560546875 ms`

eladb (approved):

PR #132, opened 2021-06-23
feat: choose submodule

### LINKED ISSUE #49: Package page - implement "choose submodule" button
There are 3 parts for this task:

### UI Element

Create a drop-down box that contains all available submodules of a package, and allows the user to search with free text.
i.e a user inputting `eks` should see the `@aws-cdk/aws-eks` submodule.

### Refresh content

once the user select a submodule, the element should pass that information to the content area, for rendering of a specific submodule. 

### Linking support

In order to support linking to a specific submodule, we will also accept a submodule name in the URL parameters of the package page. When the package is loaded, the default selection will be that coming from the URL, if none exists, no submodule is selected and the main readme of the package will be displayed.

Is a package doesn't have submodules, the component should be displayed but disabled.

---

Should be implemented after https://github.com/cdklabs/construct-hub-webapp/issues/44 since it relates to the same component, and requires some the functionality outlined their.

gabewomble:
Fixes #49 

# Summary

Implements the Choose Submodule flow for constructs with submodules. This is currently limited to submodules that are a single level deep.

Screen Shot 2021-06-23 at 1 41 06 PM
Screen Shot 2021-06-23 at 1 41 30 PM
Screen Shot 2021-06-23 at 1 42 41 PM

I prototyped a version with a modal rather than a popover which always rendered in the center of the viewport. Unfortunately the pic is cropped around the component, but the key difference is that it always renders mid-viewport, and has a dark overlay, creating a stronger contrast from the backgrounded content. Sharing below as I can't really decide which is the better UX. I believe the modal is a more a11y friendly UX, though it strays from the wireframes

Screen Shot 2021-06-22 at 5 31 26 PM


src/components/ChooseSubmodule/SearchInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const SearchInput = forwardRef<SearchInputProps, "input">(
+  ({ value, onChange, onSubmit }, inputRef) => {
+    const [inputValue, setInputValue] = useState(value);
+    const debounced = useDebounce(inputValue, 250);
+
+    useEffect(() => {
+      onChange(debounced);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [debounced]);
```
BryanPan342, line 25: Im not sure if I understand this code correctly so please let me know if I'm a totally off. But I wonder if it would be good to have a `callback` function as an option parameter for `useDebounce`.

It would solve this linter problem and I feel like most of the times with `useDebounce` there is an action that will take place after the value has been resolved (if it's different).

src/components/PackageDetails/index.tsx
```diff
@@ -18,9 +18,7 @@ interface PackageDetailsProps {
  * the Getting Started, Operator Area, and Publisher Area sections
  */
 export function PackageDetails({ assembly, metadata }: PackageDetailsProps) {
-  const isLoading = assembly.loading || metadata.loading;
-
-  if (isLoading || !metadata.data) {
+  if (metadata.loading || !metadata.data) {
```
BryanPan342, line 21: does `assembly.loading` need to be part of this conditional statement?

src/views/Package/index.tsx
```diff
@@ -23,8 +23,8 @@ export function Package() {
   const [language] = useLanguage();
 
   useEffect(() => {
-    void requestAssembly(name, version, scope);
     void requestMetadata(name, version, scope);
+    void requestAssembly(name, version, scope);
```
BryanPan342, line 27: haha if the app was next.js we could get all the info build time 😊 but all these custom hooks are very existing

BryanPan342 (commented):
just have some comments! trying to get familiar with the code base so please LMK if i get anything wrong!

src/components/PackageDetails/index.tsx
gabewomble: This was before the change Eli made for the dedicated backend, so yeah this needs to go back. (this was how I had improved load states)

src/components/ChooseSubmodule/SearchInput.tsx
gabewomble: Not a bad idea, I've updated the hook's second arg to be an options object 👍

src/views/Package/index.tsx
gabewomble: Indeed, they do exist 😛

BryanPan342: Exciting***

src/components/PackageDocs/index.tsx
```diff
@@ -55,6 +56,8 @@ export function PackageDocs({
   language,
   submodule,
 }: PackageDocsProps) {
+  const [borderColor] = useToken("colors", ["gray.100"]);
+  const border = `1px solid ${borderColor}`;
```
MrArnoldPalmer, line 60: What is the need to extract the border color token here?

Can we not just?
```
<Box borderBottom='"1px solid" borderColor="gray.100" justify="center" p={0}>
  <ChooseSubmodule assembly={assembly} />
</Box>
```

src/components/ChooseSubmodule/ChooseSubmodule.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { FunctionComponent, useCallback, useMemo, useState } from "react";
+import { useHistory, useLocation } from "react-router-dom";
+import { useQueryParams } from "../../hooks/useQueryParams";
+import { SearchPopover } from "./SearchPopover";
+
+export interface ChooseSubmoduleProps {
+  assembly?: Assembly;
+}
+
+export const ChooseSubmodule: FunctionComponent<ChooseSubmoduleProps> = ({
```
MrArnoldPalmer, line 13: If we want to move towards `const x: FunctionComponent<props> = () => {}` we should do it for everything. Keeping consistent style would be my preference for now.

src/components/ChooseSubmodule/ChooseSubmodule.tsx
```diff
@@ -0,0 +1,98 @@
...
+      results = results.filter(({ name }) =>
+        name.toLowerCase().includes(filter.toLowerCase())
+      );
+    }
+
+    return results.map(({ name }) => ({
+      name,
+      to: getUrl(name),
+    }));
+  }, [assembly?.submodules, filter, getUrl]);
```
MrArnoldPalmer, line 57: Can we compute this all and pass as props to a simple presentational instead of having this in the same function as the presentational stuff?

MrArnoldPalmer (commented):
Just some comments before ready for final review.

src/components/PackageDocs/index.tsx
gabewomble: Good catch, this is a remnant from the previous iteration which incorporated the newer design changes

src/components/ChooseSubmodule/ChooseSubmodule.tsx
gabewomble: Fair enough, figured I'd just go ahead & start moving towards that direction, but I can hold off

src/components/ChooseSubmodule/ChooseSubmodule.tsx
gabewomble: I don't really know that it makes sense to break this component apart any more than it already is

src/components/ChooseSubmodule/ChooseSubmodule.tsx
MrArnoldPalmer: lets do it all at once, and add a lint rule to prefer this going forward. do it now! but just separately.

src/components/ChooseSubmodule/ChooseSubmodule.tsx
MrArnoldPalmer: yeah I'm fine with it on further inspection.

MrArnoldPalmer (approved):

PR #133, opened 2021-06-23
chore: support dedicated backend

iliapolo:
Companion to https://github.com/cdklabs/construct-hub/pull/125

## Notable changes

- Change assembly URL according to new path.
- Changed `Metadata` type according to what is currently included in the `metadata.json` file, which is just `date` since all other information is available in the assembly.


.projenrc.js
```diff
@@ -145,7 +145,10 @@ fetchAssemblies.exec(`node scripts/fetch-assemblies.js`);
 project.gitignore.exclude("public/packages");
 
 // Proxy requests to awscdk.io for local testing
-project.package.addField("proxy", "https://awscdk.io");
+project.package.addField(
+  "proxy",
+  "https://construct-hub-testing.dev-tools.aws.dev/"
```
iliapolo, line 150: This is kind of funky. Any other ideas?

src/api/package/assemblies.ts
```diff
@@ -63,7 +63,7 @@ async function fetchAssembly(
     version = version.substring(1, version.length);
   }
 
-  const assemblyPath = `${getAssetsPath(name, version, scope)}/jsii.json`;
+  const assemblyPath = `${getAssetsPath(name, version, scope)}/assembly.json`;
```
iliapolo, line 66: https://github.com/cdklabs/construct-hub/blob/main/src/backend/shared/constants.lambda-shared.ts#L27

src/api/package/util.test.ts
```diff
@@ -7,13 +7,13 @@ describe("getAssetsUrl", () => {
 
   test("excludes scope when undefined", () => {
     const subject = getAssetsPath(name, version);
-    expect(subject).toEqual("/packages/PACKAGE_NAME@PACKAGE_VERSION");
+    expect(subject).toEqual("/data/PACKAGE_NAME/vPACKAGE_VERSION");
```
iliapolo, line 10: https://github.com/cdklabs/construct-hub/blob/main/src/backend/shared/constants.lambda-shared.ts#L12

src/api/package/util.ts
```diff
@@ -1,7 +1,7 @@
 export function getAssetsPath(name: string, version: string, scope?: string) {
-  const prefix = `/packages/`;
+  const prefix = `/data/`;
```
iliapolo, line 2: https://github.com/cdklabs/construct-hub/blob/main/src/backend/shared/constants.lambda-shared.ts#L12

src/api/package/util.ts
```diff
@@ -1,7 +1,7 @@
...
   const body = getFullPackageName(name, scope);
-  const suffix = `@${version}`;
+  const suffix = `/v${version}`;
```
iliapolo, line 4: https://github.com/cdklabs/construct-hub/blob/main/src/backend/ingestion/ingestion.lambda.ts#L70

.projenrc.js
gabewomble: Could we read this value from env or some sort of configuration file?

src/api/package/assemblies.ts
gabewomble: Could you update the corresponding service worker route as well? You can search within `service-worker.ts` for `jsii.json`

.projenrc.js
iliapolo: Mmm maybe config file we ignore. let me try.

src/api/package/assemblies.ts
iliapolo: Good catch

src/components/OperatorArea/index.tsx
```diff
@@ -22,11 +22,9 @@ export function OperatorArea({ assembly, metadata }: OperatorAreaProps) {
   const details: ReactNode[] = useMemo(() => {
     if (!metadata) return [];
 
-    const {
-      date,
-      publisher: { username },
-      links: { repository },
-    } = metadata;
+    const { date } = metadata;
+    const username = assembly?.author.name;
```
gabewomble, line 26: is this `date` field present on `assembly.metadata`? If so we can cut out the metadata.json request and just read from the assembly response

iliapolo: Nope, this is the date the package was published to npm, so not available on the assembly (which is created during package build)

.projenrc.js
iliapolo: @gabewomble This won't work since eventually it gets written to `package.json` - so we will always run the risk of committing it.

gabewomble: Actually, if this is just going to hit the same domain that it's deployed to, do we need a proxy?

iliapolo: > Forgot about local dev... the env I use the most

Ha :)

Anyway - its ok, we'll leave it like this.

src/api/package/assemblies.ts
iliapolo: Added

iliapolo:
@gabewomble Should be ready now

gabewomble (approved):

src/api/package/util.ts
eladb: Can we somehow reuse these constants across the packages?

src/service-worker.ts
```diff
@@ -90,7 +90,8 @@ registerRoute(
 // The following routes cache our jsii.json, metadata.json, and packages.json requests.
 
 registerRoute(
-  ({ url }) => url.origin === self.origin && url.pathname.endsWith("jsii.json"),
+  ({ url }) =>
+    url.origin === self.origin && url.pathname.endsWith("assembly.json"),
```
eladb, line 94: Can we move `assembly.json` to a const?

PR #137, opened 2021-06-24
feat: redirect to latest version

iliapolo:
For example: Redirects `packages/@aws-cdk/aws-eks` --> `packages/@aws-cdk/aws-eks/v/1.109.0`

As far as testing goes, there is a simple unit test to validate the url construction. I tried doing a full react rendering test but that seems a bit tricky since we need to mock a few hooks/context. @gabewomble maybe you can have a look?

In any case I think its ok for now.

Resolves https://github.com/cdklabs/construct-hub-webapp/issues/120

### DIFF
--- a/src/views/PackageLatest/PackageLatest.test.tsx
+++ b/src/views/PackageLatest/PackageLatest.test.tsx
@@ -1,0 +1,82 @@
+import { Packages } from "../../api/package/packages";
+import { buildRedirectUrl } from "./index";
+
+test("redirect to latest", () => {
+  const catalog: Packages = {
+    packages: [
+      {
+        author: {
+          name: "authorName",
+          url: "authorUrl",
+        },
+        description: "packageDescription",
+        keywords: [],
+        metadata: {
+          date: "publish date",
+        },
+        name: "my-package",
+        version: "1.109.0",
+      },
+    ],
+  };
+
+  const url = buildRedirectUrl(catalog, "my-package");
+  expect(url).toEqual("/packages/my-package/v/1.109.0");
+});
+
+test("throws if missing package", () => {
+  const catalog: Packages = {
+    packages: [
+      {
+        author: {
+          name: "authorName",
+          url: "authorUrl",
+        },
+        description: "packageDescription",
+        keywords: [],
+        metadata: {
+          date: "publish date",
+        },
+        name: "my-package",
+        version: "1.109.0",
+      },
+    ],
+  };
+
+  expect(() => buildRedirectUrl(catalog, "missing-package")).toThrow();
+});
+
+test("throws if multiple packages", () => {
+  const catalog: Packages = {
+    packages: [
+      {
+        author: {
+          name: "authorName",
+          url: "authorUrl",
+        },
+        description: "packageDescription",
+        keywords: [],
+        metadata: {
+          date: "publish date",
+        },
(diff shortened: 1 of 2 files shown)


src/views/PackageLatest/PackageLatest.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+        name: "my-package",
+        version: "1.109.0",
+      },
+    ],
+  };
+
+  expect(() => buildRedirectUrl(catalog, "missing-package")).toThrow();
+});
+
+test("throws if multiple packages", () => {
```
gabewomble, line 49: Is there any benefit to throwing if we have duplicated data (if it is valid)? We could just take the first

src/views/PackageLatest/index.tsx
```diff
@@ -7,11 +11,41 @@ interface RouteParams {
...
+  if (packages.length === 0) {
+    throw new Error(`Package ${pkg} does not exist in catalog`);
+  }
+
+  if (packages.length > 1) {
+    throw new Error(`Multiple packages found for ${pkg} in catalog`);
+  }
 
-  return <Redirect to={`${prefix}${packagePath}${suffix}`} />;
+  return packages[0];
```
gabewomble, line 50: going along with my comment from above, we could just do something like

```ts
const package = catalog.packages.find(({ name }) => name === pkg);

if (!package) {
  throw new Error(...);
}

return package;
```

src/views/PackageLatest/PackageLatest.test.tsx
iliapolo: I consider duplicate data to be invalid, means something is wrong somewhere. A safer bet would be to throw IMO.

src/views/PackageLatest/index.tsx
iliapolo: https://github.com/cdklabs/construct-hub-webapp/pull/137#discussion_r658056724

src/views/PackageLatest/PackageLatest.test.tsx
gabewomble: We could log an error, but from a user perspective, if we have data, wouldn't it be better to display it rather than crash the app?

iliapolo: I see what you're saying and I would be more inclined to agree if this would have prevented displaying other reliable data (for example if this would have crashed search results or something).

In this case, it only breaks displaying a package who's data cannot be trusted. The contract of the catalog is that it only contains the latest version, so in theory the data structure should have been a map, and then we wouldn't be having this (synthetic) problem in the code.

Question is - Is it better to display unreliable data than to display nothing? My backend engineering history says err on the side of caution.

I say we let @MrArnoldPalmer be the tie breaker :)

MrArnoldPalmer: I think I'm with Eli here, this is a scenario where data is obviously bad and we don't have a way to decide which of the duplicates we should trust if any.  Crashing is still obviously not ideal, but in the future we can show a  'sorry, something went wrong, go somewhere else' type message.

Since there are currently numerous crash cases that should be prettier in the app, I wouldn't block on this now.

MrArnoldPalmer (approved):

PR #140, opened 2021-06-24
feat: choose submodule as a modal

gabewomble:
Updates the Choose Submodule Popover to be a modal. This is to provide a better contrast from the background content and provide a more mobile-friendly UX. This could eventually be used as a pattern for search / filter components in our app

Screen Shot 2021-06-24 at 4 07 40 PM
Screen Shot 2021-06-24 at 4 08 18 PM


src/components/ChooseSubmodule/SearchModal.tsx
```diff
@@ -0,0 +1,94 @@
+import {
```
gabewomble, line 1: This is just the old `SearchPopover.tsx` except `Popover` became `Modal` across the board, and the `ModalContent` needed to be wrapped in a `ModalOverlay`

MrArnoldPalmer (approved):

Issue #139, opened 2021-06-24
Update the search algorithm to filter by programming language and CDK type


addihorowitz:
We should support programming language and CDK type filters

gabewomble:
Note: we currently do not have a way to determine cdk type on the front end

PR #146, opened 2021-06-28
feat: catalog search component

gabewomble:
Prerequisite to #139 

This PR implements a responsive search component with state controlled externally and a language filter. Once we have the capability to filter on CDK type, a small update will be needed to add that functionality. This component is tested in storybook and unit tests, and will be integrated into the app in a follow-up to reduce PR size.

This work also co-locates all component stories with their respective components

Screen Shot 2021-06-28 at 5 22 02 PM
Screen Shot 2021-06-28 at 5 22 07 PM


.storybook/main.js
```diff
@@ -8,11 +8,11 @@ module.exports = {
   ],
   addons: [
     "@storybook/addon-a11y",
-    "@storybook/addon-essentials",
     "@storybook/addon-links",
     "@storybook/addon-storysource",
     "@storybook/preset-create-react-app",
-    "storybook-addon-performance/register"
+    "storybook-addon-performance/register",
+    "@storybook/addon-essentials",
```
gabewomble, line 15: Per storybook docs, for the actions addon (bundled with addon-essentials) to be guaranteed to work with other addons it should be the last addon loaded to take higher precedence

gabewomble:
Force-pushed an update to revert changes that are unscoped to this PR

src/components/CatalogSearch/CatalogSearch.tsx
```diff
@@ -0,0 +1,112 @@
...
+}) => {
+  const handleQueryChange: ChangeEventHandler<HTMLInputElement> = (e) => {
+    e.preventDefault();
+    onQueryChange(e.target.value);
+  };
+
+  const handleSubmit: FormEventHandler<HTMLFormElement> = (e) => {
+    e.preventDefault();
+    onSubmit({ query, language });
+  };
```
MrArnoldPalmer, line 77: When accepting event handlers as props, I think we should just let the parent specify whether `preventDefault` is called or not and just accept the top level event handler function.

IE
```ts
<Input onChange={props.onChange} />
```
This makes the components more generic and allows callers more control over inputs.

Additionally, `onSubmit` accepts an object passing query and language, both of which are values controlled by the parent. There is no need for this component to know about the structure of this function and the values it accepts. Once again `onSubmit={props.onSubmit}` and the parent can pass the values where they need to go. This decouples the two significantly imo.

MrArnoldPalmer (approved):
Suggestion on reformatting event handler props is non-blocking.

PR #149, opened 2021-06-29
chore: render date/timestamps using the <time> element

RomainMuller:
Render the complete information using the `<time>` element, instead of
losing the resolution. This additionally may make the content more
accessible to screen readers.

### DIFF
--- a/src/components/OperatorArea/OperatorArea.tsx
+++ b/src/components/OperatorArea/OperatorArea.tsx
@@ -33,7 +33,14 @@ export const OperatorArea: FunctionComponent<OperatorAreaProps> = ({
 
     if (date) {
       const publishDate = new Date(date);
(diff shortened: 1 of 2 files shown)


src/components/OperatorArea/OperatorArea.tsx
```diff
@@ -33,7 +33,14 @@ export const OperatorArea: FunctionComponent<OperatorAreaProps> = ({
 
     if (date) {
       const publishDate = new Date(date);
-      items.push(`Published: ${format(publishDate, "MMMM dd, yyyy")}`);
+      items.push(
```
gabewomble, line 36: We should eventually implement a simple `<Time />` component so we can make this consistent!

gabewomble (approved):
Nice 😎

PR #150, opened 2021-06-29
chore: change website title to "Construct Hub"

eladb:
This has been driving me crazy :-)


gabewomble (approved):
We'll eventually want to bring in `react-helmet` or a similar library so we can setup relevant meta for different pages, but this was definitely way overdue 😅

PR #151, opened 2021-06-29
feat: integrate catalog search with filters

### LINKED ISSUE #139: Update the search algorithm to filter by programming language and CDK type

gabewomble:
Fixes #139 

### Key Features

This PR implements the <CatalogSearch /> component on the search results page, allowing users to filter by supported language. Because searching is used across multiple views in our application with separate components (header search and catalog search), I've also introduced two new hooks which will encapsulate search logic:

`useCatalogSearch`: Exposes searchQuery and language state, handlers to update the state, and a submit method
`useCatalogResults`: Exposes a stateful object which returns filtered search results as well as displayable results based on `limit` and `offset`

By encapsulating common search logic to these hooks, we can keep our components dumber / presentational and expose re-usable business logic

### Additional Chanegs
- begins to setup a theme for our application (currently just updates blue palette)
- adjusts search layout to be more in-line with designs
- constants for app routes, api paths, and query params

Screen Shot 2021-06-29 at 4 24 51 PM
Screen Shot 2021-06-29 at 4 25 28 PM


gabewomble:
Leaving as draft to prevent auto-merge into parent branch

MrArnoldPalmer (approved):

PR #155, opened 2021-06-30
feat: dropdown button for getting started

### LINKED ISSUE #51: Main Header --> Getting Started
Implement drop-down menu from the Menu Header. When the user clicks one of the items, a new tab should be opened.

Here are the links for getting started drop-down menu:
CDK
* [Quick Start by the community](https://cdkworkshop.com/)
* [Developer Guide](https://docs.aws.amazon.com/cdk/latest/guide/getting_started.html)
* [CLI](https://docs.aws.amazon.com/cdk/latest/guide/cli.html)
* [Best Practices](https://aws.amazon.com/blogs/devops/best-practices-for-developing-cloud-applications-with-aws-cdk/)
* [Community Slack Channel](https://cdk.dev/)

CDK8s
* [Getting Started](https://cdk8s.io/)

CDKtf
* [Getting Started](https://learn.hashicorp.com/tutorials/terraform/cdktf)

BryanPan342:
Create a dropdown button for the getting started page!

I decided to leave the styling as is. I think we can resolve most of our ugliness (esp w/ the `Browse`) when we turn `Browse` into a button and anchor search to the left.

--- 

Screen Shot 2021-06-30 at 6 34 56 PM

Screen Shot 2021-06-30 at 6 35 02 PM

Fixes #51


src/components/GettingStartedDropdown/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+GettingStartedButton.displayName = "GettingStartedButton";
+
+export default function GettingStartedDropdown(): JSX.Element {
+  return (
+    <Menu>
+      <MenuButton
+        as={GettingStartedButton}
+        rightIcon={<ChevronDownIcon color="black.600" h={6} w={6} />}
+      >
```
gabewomble, line 36:
```suggestion
      <MenuButton
        as={Button}
        bg="white"
        size="md"
        rightIcon={<ChevronDownIcon color="black.600" h={6} w={6} />}
      >
```

I know I did the same thing as this on the Dependency Dropdown, but it turns out it wasn't necessary at all 😅

src/components/GettingStartedDropdown/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  );
+}
+
+const GettingStartedButton = forwardRef(
+  (props: PropsOf<typeof Button>, ref) => {
+    return <Button bg="white" ref={ref} size="md" {...props} />;
+  }
+);
+
+GettingStartedButton.displayName = "GettingStartedButton";
```
gabewomble, line 28: This can be removed, see comment below

src/components/GettingStartedDropdown/index.tsx
BryanPan342: Ah is this cuz header is used outside the routing?

gabewomble:
Code looks pretty nice. I agree will eventually want to standardize our dropdown patterns.
Would like to see the following:
- A generic `<ExternalLInk />` which composes the chakra-ui `<Link />`. That way we have a component which adheres to the styled system and is theme-aware and handles noreferrer / target _blank
- Screenshots of the work on your PR 🙂

src/components/GettingStartedDropdown/index.tsx
gabewomble: We resolved this conversation offline, but the gist is that we can just pass in the chakra `<Button />` and `<MenuButton />` will infer the correct prop typings, so defining a separate forwardRef component to pre-define some props isn't needed

src/components/GettingStartedDropdown/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  MenuGroup,
+  MenuItem,
+  MenuList,
+} from "@chakra-ui/react";
+import { GETTING_STARTED } from "../../constants/links";
+import { ExternalLink } from "../ExternalLink";
+
+export default function GettingStartedDropdown(): JSX.Element {
+  return (
+    <Menu colorScheme="#002954">
```
gabewomble, line 16: Couple of things:
- The pattern we're using for files is to name components as `Component.tsx` & use an `index.ts`
- We've been preferring named exports rather than default
- We should be using colors from theme as much as possible rather than a hex value

BryanPan342: How do i find the themes we use?

src/components/GettingStartedDropdown/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <MenuButton
+        as={Button}
```
gabewomble, line 18: The button should be probably be using the `link` variant - https://chakra-ui.com/docs/form/button#button-variants

src/components/GettingStartedDropdown/index.tsx
BryanPan342: nvm i found it!

src/components/ExternalLink/index.tsx
```diff
@@ -0,0 +1,5 @@
+import { Link, PropsOf } from "@chakra-ui/react";
+
+export function ExternalLink(props: PropsOf<typeof Link>) {
```
gabewomble, line 3: While this ExternalLink does share the same props as the base Link, it still needs to be composed with `forwardRef` from `@chakra-ui/react` so that we can pass it to the `as` prop i.e `<Button as={ExternalLink} />`

[Here's](https://github.com/cdklabs/construct-hub-webapp/blob/0d9aaee3bfaa086941dacca7beacd72bda846336/src/components/ExternalLink/ExternalLink.tsx) an implementation you can refer to (needed an external link for one of my feature branches as well)

Additionally, we should spread props after the default prop settings so that we invert control to the developer rather than the component

src/components/GettingStartedDropdown/index.tsx
BryanPan342: Why is this a link variant? The figma shows a button as the component

gabewomble: Are we looking at the same one? 😅 
Screen Shot 2021-07-01 at 10 28 08 AM

gabewomble (changes requested):
Left a few comments

gabewomble (approved):
Looks good! Mergify merged the <ExternalLink /> from my branch in, but it's identical to yours so up to on which one to keep 🙂

PR #164, opened 2021-07-01
feat: use construct button

### LINKED ISSUE #121: Implement Use Construct Button
When Use Construct is pressed a drop-down modal should be displayed with installation instructions:
* For python: pip install <package_name>
* For TypeScript: npm install <package_name>

gabewomble:
Fixes #121

This PR implements the Use Construct button using the `<CodePopover />` and `<CodePopoverTrigger />` components from the parent branch. 

Additional Changes:
- colocates Package Page components with the Package page, rather than general components (finally)
- Updates the language bar & language selection to the newer design

Screen Shot 2021-07-01 at 4 32 58 PM
Screen Shot 2021-07-01 at 4 33 15 PM


src/components/LanguageBar/LanguageBar.tsx
```diff
@@ -55,53 +55,57 @@ export const LanguageBar: FunctionComponent<LanguageBarProps> = ({
...
   return (
-    <Flex data-testid="language-bar">
-      {(showDisabled ? LANGUAGES : targetLanguages).map(
-        (language: Language) => {
-          const isDisabled = !targetLanguages.includes(language);
-          const isSelected = language === selectedLanguage;
+    <Grid
+      data-testid="language-bar"
+      gap="2"
+      gridTemplateColumns={`repeat(${displayable.length}, 1fr)`}
```
gabewomble, line 63: The advantage of using grid is that it forces each item to have the exact same width

src/views/Package/components/ChooseSubmodule/SearchInput.tsx
```diff
@@ -6,8 +6,8 @@ import {
   forwardRef,
 } from "@chakra-ui/react";
 import { FormEventHandler, useState } from "react";
-import { useDebounce } from "../../hooks/useDebounce";
-import { Form } from "../Form";
+import { Form } from "../../../../components/Form";
```
gabewomble, line 9: Will love having absolute paths so we can just import from "components/Form"

MrArnoldPalmer (approved):

PR #165, opened 2021-07-02
feat: cookie consent

gabewomble:
Implements the cookie consent API via a set of wrapping functions and a context which exposes a public API after initializing the cookie consent API on mount


src/lib/shortbread/shortbread.ts
```diff
@@ -0,0 +1,97 @@
...
+  language?: Navigator["language"];
+  __storeWriter?: (cookie: Document["cookie"]) => void;
+}
+
+export interface Shortbread {
+  /**
+   * Call this function to be notified when you cookie category is allowed
+   */
+  access: (
+    name: "essential" | "performance" | "functional" | "advertising",
```
RomainMuller, line 15: Could this use the `enum` instead?

src/constants/cookie-categories.ts
```diff
@@ -0,0 +1,6 @@
+export enum CookieCategories {
```
RomainMuller, line 1: This enum isn't used anywhere... this is weird?

gabewomble: Ahh yeah, I had some additional work on this branch that used this enum, but after syncing with Mitch, decided to cut it. I'll just drop this enum for now 👍

MrArnoldPalmer (approved):

PR #166, opened 2021-07-02
feat: Expandable / Collapsible Nav tree

### LINKED ISSUE #162: Implement functionality to collapse / expand Package Nav items
The current state of the Package Nav is non-interactive, meaning that the links function as links, but we need to support the functionality to collapse & expand items as-needed.

User Interactions:
- Expandable / Collapsible items should have a chevron icon that allows their open state to be toggled without navigating to their href
- Clicking on an item's text will always navigate / scroll to the href
- Active NavItems should be expanded by default

gabewomble:
Fixes #162

This PR implements a Nav Tree component which supports the following functionality:
- Collapsing and expanding items with children
- Automatically expanding the path of an active item

The way this component manages to open all parents of an active item is via the following strategy.
1. All parents pass their `onOpen` to their children.
2. All items which receive `props.onOpen` wrap their own `onOpen` with the `props.onOpen`
3. All children are rendered, even if hidden, allowing lifecycle hooks to run
4. When a child mounts, if the window's url matches the child's url, it will call the wrapped onOpen, implicitly opening all parent trees

https://user-images.githubusercontent.com/8749859/124321524-72306c80-db43-11eb-9a0d-b7aefda22ed9.mov

Screen Shot 2021-07-06 at 9 40 06 AM


gabewomble:
working on improving some styles

MrArnoldPalmer (approved):

PR #167, opened 2021-07-02
chore: fix mapped keys for <GettingStartedDropdown />

gabewomble:
This is a silly and harmless bug (which logs an error in console) from the GettingStartedDropdown. Mapped items need a key at the root jsx element, even if the element is a fragment 🙂


BryanPan342 (approved):
Thanks for the catch! Totally forgot I had a conditional statement for the divider

PR #178, opened 2021-07-08
feat: 404 page

gabewomble:
Adds a 404 page design!

Screen Shot 2021-07-08 at 10 43 16 AM


src/components/Picture/Picture.tsx
```diff
@@ -0,0 +1,21 @@
+import { Box, HTMLChakraProps, forwardRef, Image } from "@chakra-ui/react";
+
+export interface PictureProps extends HTMLChakraProps<"img"> {
```
gabewomble, line 3: This extends the "img" b/c we want to forward styles to the `<Image />`. We still pass the ref to the root element, which renders a `<picture />`

MrArnoldPalmer (approved):
Looks good!

Added https://github.com/cdklabs/construct-hub-webapp/issues/179 for future work if we decide we want it.

PR #183, opened 2021-07-12
chore: fetch markdown from backend 

iliapolo:
Following https://github.com/cdklabs/construct-hub/pull/162, we should now be fetching the markdown from the backend instead of generating it in the client. 

In addition, switch usages of `reflect.Assembly` (i.e `jsii-reflect`) to `spec.Assembly` (i.e `@jsii/spec`) since it is not required anymore and `spec.Assembly` is an easier type to work with.


src/api/package/assembly.ts
```diff
@@ -0,0 +1,20 @@
+import * as spec from "@jsii/spec";
+import { API_PATHS } from "../../constants/url";
+import { getAssetsPath } from "./util";
+
+export const fetchAssembly = async (
```
eladb, line 5: Docstring

iliapolo: Added

src/api/package/metadata.ts
```diff
@@ -8,6 +8,9 @@ export interface Metadata {
   };
 }
 
+/**
+ * Fetch metadata of a specific package from the backend.
+ */
```
gabewomble, line 13: I believe there's an eslint rule that enforces jsdoc / docstrings. May be worth adding in a future PR so we can enforce this?

src/constants/languages.ts
```diff
@@ -7,12 +7,18 @@ import { NodeIcon } from "../icons/NodeIcon";
 import { PythonIcon } from "../icons/PythonIcon";
 import { TSIcon } from "../icons/TSIcon";
 
-export type Language = "dotnet" | "js" | "ts" | "python" | "golang" | "java";
+export type Language =
+  | "dotnet"
+  | "js"
+  | "typescript"
```
gabewomble, line 13: 👍

src/views/Package/Package.tsx
```diff
@@ -1,7 +1,8 @@
 import { Box, Stack } from "@chakra-ui/react";
 import { FunctionComponent, useEffect } from "react";
 import { useParams } from "react-router-dom";
-import { createAssembly } from "../../api/package/assemblies";
+import { fetchAssembly } from "../../api/package/assembly";
+import { fetchMarkdown } from "../../api/package/docs";
```
gabewomble, line 5: Non-blocking: might be worth adding an index file to `api/package` so we can shorten the imports?

```ts
import { fetchAssembly, fetchMarkdown, fetchMetadata } from "../../api/package";
```

src/views/Package/Package.tsx
```diff
@@ -46,13 +51,15 @@ export const Package: FunctionComponent = () => {
...
       </Box>
       {/* Readme and Api Reference Area */}
-      {assemblyResponse.data && !assemblyResponse.loading && (
-        <PackageDocs
-          assembly={assemblyResponse.data}
-          language={language}
-          submodule={q.get(QUERY_PARAMS.SUBMODULE) ?? ""}
-        />
-      )}
+      {markdownResponse.data &&
```
gabewomble, line 54: Another non-blocker: I wonder if it would make sense to add a property to `useRequest`'s return value such as `loaded` so we can simplify these conditions

src/views/Package/components/ChooseSubmodule/ChooseSubmodule.tsx
```diff
@@ -43,18 +43,21 @@ export const ChooseSubmodule: FunctionComponent<ChooseSubmoduleProps> = ({
   );
 
   const submodules = useMemo(() => {
-    let results = assembly?.submodules ?? [];
+    let results = Object.keys(assembly?.submodules ?? {});
```
gabewomble, line 46: Nit, but we could set a const in the component's scope for submodule names - 

`const submoduleNames  = Object.keys(assembly?.submodules ?? {});`

so that we can re-use this for the `disabled` prop in favor of repeating the `Object.keys` logic

src/views/Package/components/PackageDocs/Body.tsx
```diff
@@ -72,13 +73,21 @@ const components = ChakraUIRenderer({
...
   const body = useMemo(() => {
-    return (
-      <ReactMarkdown components={components} rehypePlugins={rehypePlugins}>
-        {children}
-      </ReactMarkdown>
-    );
-  }, [children]);
+    if (format === "raw") {
+      return <div>{children}</div>;
+    }
```
gabewomble, line 81: imo we can probably cut this render format since it was really just for debugging / identifying perf issues

src/views/Package/components/PackageDocs/PackageDocs.tsx
```diff
@@ -46,34 +43,11 @@ export const appendItem = (itemTree: Item[], item: Element): Item[] => {
...
-      language: language,
-      submoduleName: submodule,
-    });
-
-    const md = doc.render();
-    const s = md.render();
-    console.timeEnd(timeLabel);
-    return s;
-  }, [hasApiReference, assembly, language, submodule]);
+  const source = markdown;
```
gabewomble, line 50: nitpick, but this alias seems unnecessary

gabewomble:
Looks great on a first pass, really excited for this feature!

src/views/Package/Package.tsx
```diff
@@ -19,16 +20,20 @@ interface PathParams {
...
   const [language] = useLanguage();
+  const submodule = q.get(QUERY_PARAMS.SUBMODULE) ?? "";
 
   useEffect(() => {
     void requestMetadata(name, version, scope);
+    void requestMarkdown(name, version, language, scope, submodule);
     void requestAssembly(name, version, scope);
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [name, scope, version]);
+  }, [name, scope, version, language]);
```
gabewomble, line 36: Follow-up from standup:
> Changing Submodule does not refetch the data

The solution we should implement here would be to add a separate `useEffect` specifically for the markdown which will subscribe to changes to the `language` and `submodule`

```tsx
useEffect(() => {
  void requestMarkdown(name, version, language, scope, submodule);
}, [name, scope, version, language, submodule]);
```

The reason we would want this to be a separate effect is b/c we do not need to request the assembly or metadata whenever language or submodules change since assembly & metadata information will be the same regardless

src/views/Package/Package.tsx
```diff
@@ -19,16 +20,20 @@ interface PathParams {
...
+  const [requestAssembly, assemblyResponse] = useRequest(fetchAssembly);
   const [requestMetadata, metadataResponse] = useRequest(fetchMetadata);
+
   const q = useQueryParams();
   const [language] = useLanguage();
+  const submodule = q.get(QUERY_PARAMS.SUBMODULE) ?? "";
 
   useEffect(() => {
     void requestMetadata(name, version, scope);
+    void requestMarkdown(name, version, language, scope, submodule);
```
gabewomble, line 33: also nit, but I think we should shift these requests to use named parameters

src/api/package/metadata.ts
iliapolo: Yeah, i'll take a note but leave it for now.

src/views/Package/Package.tsx
iliapolo: Cool thanks. Done 👍

src/api/package/docs.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const markdownPath = `${getAssetsPath(name, version, scope)}${docsSuffix}`;
+  const response = await fetch(markdownPath);
+  if (!response.ok) {
+    throw new Error(
+      `Failed fetching documentation for ${markdownPath}: ${response.statusText}`
+    );
+  }
+  return response.text();
+};
```
RomainMuller, line 23:
```suggestion
export async function fetchMarkdown(
  name: string,
  version: string,
  language: string,
  scope?: string,
  submodule?: string,
): Promise<string> {
  const docsSuffix = `/docs-${submodule ? `${submodule}-` : ""}${language}.md`;

  const markdownPath = `${getAssetsPath(name, version, scope)}${docsSuffix}`;
  const response = await fetch(markdownPath);
  if (!response.ok) {
    throw new Error(
      `Failed fetching documentation for ${markdownPath}: ${response.statusText}`
    );
  }
  return response.text();
}
```

src/views/Package/components/ChooseSubmodule/ChooseSubmodule.tsx
```diff
@@ -77,7 +80,7 @@ export const ChooseSubmodule: FunctionComponent<ChooseSubmoduleProps> = ({
         borderRadius="none"
         color="blue.500"
         data-testid="choose-submodule-search-trigger"
-        disabled={!assembly?.submodules.length}
+        disabled={!Object.keys(assembly?.submodules ?? {}).length}
```
RomainMuller, line 83:
```suggestion
        disabled={Object.keys(assembly?.submodules ?? {}).length === 0}
```

RomainMuller (approved):
LGTM, besides I don't understand why so much craze over the arrow-functions... In places where a regular function would be fine...

PR #185, opened 2021-07-13
feat: json rendering

MrArnoldPalmer:
Fixes #

### DIFF
--- a/src/api/docgen/view/documentation.ts
+++ b/src/api/docgen/view/documentation.ts
@@ -1,4 +1,5 @@
 import * as reflect from "jsii-reflect";
+import { PackagePageContent } from "../api-reference";
 import { Markdown } from "../render/markdown";
 import { PythonTranspile } from "../transpile/python";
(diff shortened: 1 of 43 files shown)


iliapolo:
@eladb

src/api/docgen/view/documentation.ts
```diff
@@ -100,6 +101,25 @@ export class Documentation {
...
+    const apiReference = new ApiReference(
+      this.transpile,
+      this.assembly,
+      this.submodule
+    );
+
+    return {
+      readme,
+      apiReference: apiReference.renderToJson(),
+      navItems: [],
```
gabewomble, line 119: Are there plans for this class to generate navItems? Not sure why we would return them here otherwise (unless it is placeholder)

MrArnoldPalmer: placeholder for now, we will derive these from the api reference json + readme me on the client.

iliapolo:
keep

PR #193, opened 2021-07-13
feat: Markdown styling & syntax highlighting

gabewomble:
This PR is a first pass at styling our markdown

The bulk of additions are setting basic colors / spacing for primitives, however we now have 3 separate ways to render code:
1. Inline - this is a basic code primitive with no highlighting
2. MultiLine - this is for codeblocks and displays a top bar with a copy icon
3. SingleLine - similar to codeblocks but displays the copy icon inline (not sure if we want to keep this one, but the multiLine renderer for codeblocks with a single line looked excessive)

Screenshot_2021-07-13 Construct Hub

Screenshot_2021-07-13 Construct Hub(1)

Screenshot_2021-07-13 Construct Hub(2)


src/components/Code/Code.tsx
```diff
@@ -0,0 +1,125 @@
...
+    color: "white",
+    sx: {
+      ":hover": {
+        bg: "blue.400",
+      },
+    },
+  },
+};
+
+const CopyButton: FunctionComponent<{
```
gabewomble, line 33: I'll probably generalize this component this component in a later commit, this PR is in draft since it still needs a bit of cleanup as some stuff is still slightly rushed

src/components/Code/Code.tsx
```diff
@@ -0,0 +1,125 @@
...
+      minW="auto"
+      onClick={onCopy}
+      variant="ghost"
+      w={6}
+      {...theme}
+    />
+  );
+};
+
+const SingleLineRenderer: FunctionComponent<RenderProps & { code: string }> = ({
```
gabewomble, line 54: These renderers will probably also get their own files, and the syntax highlighting code will become a general component

src/components/CopyButton/CopyButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+        ref={ref}
+        variant="ghost"
+        w={6}
+        {...btnProps}
+      />
+    );
+  }
+);
+
+CopyButton.displayName = "CopyButton";
```
MrArnoldPalmer, line 35: Are displaynames not forwarded automatically when wrapped with forwardRef?

src/components/Markdown/Code.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const code = Children.toArray(children)
+    .reduce((accum: string, child): string => {
+      if (typeof child === "string") {
+        return `${accum}${child}`;
+      }
+      return accum;
+    }, "")
+    .trim();
+
+  return <BlockCode code={code} language={"python"} />;
```
MrArnoldPalmer, line 40: Do we want the useLanguage hook here? I assume this was for testing as it's commented out above.

src/components/Markdown/Markdown.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { FunctionComponent } from "react";
+import ReactMarkdown from "react-markdown";
+import rehypeRaw from "rehype-raw";
+import { Code } from "./Code";
+import { Headings } from "./Headings";
+import { Hr } from "./Hr";
+import { Ul, Ol, Li } from "./List";
+import { A, Blockquote, Em, P, Pre, Sup } from "./Text";
+
+const components = ChakraUIRenderer(
```
MrArnoldPalmer, line 12: Are we fully overriding everything? If so I think we can remove this renderer all together.

MrArnoldPalmer (commented):
Some general questions, no explicit changes requested.

src/components/CopyButton/CopyButton.tsx
gabewomble: I thought so, but in dev tools they show up as `Anonymous`. Will make a follow-up to update the other forwardRef components

src/components/Markdown/Code.tsx
gabewomble: This was some carry-over from your implementation, but realistically - no. We should really only use it for the api reference section so it'll need to be passed in as a prop

src/components/Markdown/Markdown.tsx
gabewomble: I'll check, I've left it in to be safe, but there's always an edge case 🙃

src/components/Markdown/Code.tsx
gabewomble: Actually gonna pass it in for now - no harm in having mostly correct syntax highlighting

src/components/Markdown/Markdown.tsx
gabewomble: Ok it's good 2 remove

src/components/CopyButton/CopyButton.tsx
gabewomble: Added to other forwardRef components

MrArnoldPalmer (approved):

PR #225, opened 2021-07-19
feat: better configure ReactMarkdown

RomainMuller:
Enable remark-gfm so that all GitHub-Flag Markdown is supported,
including tables (Fixes cdklabs/construct-hub#197).

Properly sanitize all generated markdown using `rehype-sanitize`,
ensuring we do not render anything that GitHub would not.

Transform image and link URLs that are repository-relative, so that such
links are valid. These resolve to GitHub distribution points in the same
way that npmjs.com does, so this change is neutral for packages that are
not configured with a GitHub repository (Fixes cdklabs/construct-hub#198,
fixes cdklabs/construct-hub#199).

### DIFF
--- a/.projen/deps.json
+++ b/.projen/deps.json
@@ -234,6 +234,14 @@
       "type": "runtime"
     },
     {
+      "name": "rehype-sanitize",
+      "type": "runtime"
+    },
+    {
+      "name": "remark-gfm",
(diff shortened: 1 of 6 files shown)


eladb:
YASSS! So much goodness!

gabewomble (approved):

PR #237, opened 2021-07-21
fix: headings are not prominent enough

iliapolo:
This PR fixes:

- List items were not aligned with the text (they appeared more on the left). This was solved by removing any explicit margin definitions and letting the markdown component handle it.
- Headings which were formatted with `pre` (i.e properties/parameters) received a smaller font size than their appropriate h-level since this is what the `Code` element does. Resolve by applying special CSS selector to override this behavior. Also changed to color to black and reserve blue colors to links only.
- Increase grey background for those headings to create a clear distinction between them and other headings. 
- Fix list item prefixes in metadata to be *italic*, as originally intended and decrease font size of metadata.

### Before

Screen Shot 2021-07-20 at 11 42 28 PM
 
### After

Screen Shot 2021-07-20 at 11 41 57 PM

Fixes https://github.com/cdklabs/construct-hub/issues/201

### DIFF
--- a/src/components/Markdown/Headings.tsx
+++ b/src/components/Markdown/Headings.tsx
@@ -49,7 +49,7 @@ export const Headings: FunctionComponent<HeadingResolverProps> = ({
         data-heading-title={title}
         id={id}
         replace
-        sx={{ "> code": { color: "black", fontSize: "inherit" } }}
+        sx={{ "> code": { color: "blue.800", fontSize: "inherit" } }}
         to={`#${id}`}
       >
         {children}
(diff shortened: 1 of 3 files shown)


src/components/Markdown/Headings.tsx
```diff
@@ -49,6 +49,7 @@ export const Headings: FunctionComponent<HeadingResolverProps> = ({
         data-heading-title={title}
         id={id}
         replace
+        sx={{ "> code": { color: "black", fontSize: "inherit" } }}
```
gabewomble, line 52: IMO we should keep the color within our theme & consistent with other headings which is `blue.800`

iliapolo: Fixed 👍

gabewomble (approved):

PR #240, opened 2021-07-21
feat: external link warning modal

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-internal/issues/30

Implements a warning modal for external links (specifically enabled for user-generated links)
The flow is as follows:
1. User content external links warn by default
2. Proceeding opens the external link in a new tab
3. Cancelling closes the warning and does not open the link
4. If a user updates the preference to not show in the future, the change is only applied if they click proceed

https://user-images.githubusercontent.com/8749859/126523011-7b8812a8-2664-4c50-a362-4d1dc7d35504.mov

### DIFF
--- a/src/components/ExternalLink/ExternalLink.tsx
+++ b/src/components/ExternalLink/ExternalLink.tsx
@@ -1,15 +1,27 @@
 import { ExternalLinkIcon } from "@chakra-ui/icons";
 import { forwardRef, Link, LinkProps } from "@chakra-ui/react";
+import { useExternalLinkWarning } from "../../contexts/ExternalLinkWarning";
 
 export interface ExternalLinkProps extends LinkProps {
+  /**
+   * Shows an external link icon. `true` by default
+   */
   hasIcon?: boolean;
+  /**
+   * Prompts the user to confirm leaving the site. `false` by default
+   */
+  hasWarning?: boolean;
 }
 
 export const ExternalLink = forwardRef<ExternalLinkProps, "a">(
-  ({ hasIcon = true, children, ...props }, ref) => {
+  ({ children, hasIcon = true, hasWarning, href, onClick, ...props }, ref) => {
+    const withPrompt = useExternalLinkWarning();
+
     return (
       <Link
         color="blue.500"
+        href={href}
+        onClick={hasWarning ? withPrompt({ href, onClick }) : onClick}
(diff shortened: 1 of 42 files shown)


src/contexts/ExternalLinkWarning/ExternalLinkWarning.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { ExternalLinkWarningModal } from "./ExternaLinkWarningModal";
+
+const ExternalLinkWarningContext = createContext<ExternalLinkPrompt>(
+  ({ onClick }) => onClick
+);
+
+export const useExternalLinkWarning = () =>
+  useContext(ExternalLinkWarningContext);
+
+export const ExternalLinkWarningProvider: FunctionComponent = ({
```
gabewomble, line 24: This component is setup as it's own context as we don't want every Link to instantiate a unique modal in the dom. Instead, all external links can re-use the same instance and can pass an onclick & href to the modal when they click an external link

src/components/ExternalLink/ExternalLink.tsx
```diff
@@ -1,15 +1,27 @@
...
 import { forwardRef, Link, LinkProps } from "@chakra-ui/react";
+import { useExternalLinkWarning } from "../../contexts/ExternalLinkWarning";
 
 export interface ExternalLinkProps extends LinkProps {
+  /**
+   * Shows an external link icon. `true` by default
+   */
   hasIcon?: boolean;
+  /**
+   * Prompts the user to confirm leaving the site. `false` by default
```
iliapolo, line 11: Why default to `false`? Feels like `true` is the safer choice.

src/components/Markdown/Headings.tsx
```diff
@@ -52,7 +51,6 @@ export const Headings: FunctionComponent<HeadingResolverProps> = ({
         to={`#${id}`}
       >
         {children}
-        {level < 3 && <Hr />}
```
iliapolo, line 55: What's this?

gabewomble: I noticed whenever user markdown already has dividers between headers, this would create an extra one. IMO it'll be better if we just update the API reference markdown to place dividers as needed

iliapolo (approved):

PR #241, opened 2021-07-21
feat: report package link

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-internal/issues/1

Also adds package version to the Package Page header

Screen Shot 2021-07-21 at 1 50 25 PM
Screen Shot 2021-07-21 at 1 51 56 PM


src/views/Package/components/OperatorArea/OperatorArea.tsx
```diff
@@ -74,6 +81,12 @@ export const OperatorArea: FunctionComponent<OperatorAreaProps> = ({
       items.push(<>Registry: {registryLink}</>);
     }
 
+    items.push(
```
iliapolo, line 84: The location of the link feels out of place to me. It's different than other links we show here and I think deserves a place of its own. This also makes it hard to notice it. 

Can we move to the top right corner of the panel?

src/views/Package/components/OperatorArea/OperatorArea.tsx
```diff
@@ -74,6 +81,12 @@ export const OperatorArea: FunctionComponent<OperatorAreaProps> = ({
...
+      <Link color="blue.500" href="mailto:abuse@amazon.com">
```
iliapolo, line 85: Maybe prefill the subject? Will make it easy for us to filter those out later.  

```suggestion
      <Link color="blue.500" href="mailto:abuse@amazon.com?subject={`ConstructHub - Report of abusive package: ${package}`}">
```

iliapolo (changes requested):
Marking as changes requested to better filter out PR's that still need a review.

src/views/Package/components/OperatorArea/OperatorArea.tsx
gabewomble: I can move it to the top, but placing it top right seems out of place

iliapolo (approved):

PR #250, opened 2021-07-22
chore: remove duplicate build tasks

Chriscbr:
This removes some unnecessary extra build tasks, which should reduce build times on GitHub actions. (If you look at the `.projen/tasks.json`, the projen/compile/test steps were being run twice).


gabewomble (approved):
Nice 😎

Issue #9, opened 2021-04-21
Implement: Analytics


RomainMuller:
Implement the analytics solution designed as part of #8, and roll it out to production.

PR #251, opened 2021-07-22
feat: pageview analytics

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-webapp/issues/9

This draft implements pageview analytics for our application. When a user navigates to a page, an analytics call will be made which capture the page name, url, and event name. This PR is going into draft as it will need reviews from some of the EU folks who are away until monday.

As for the implementation:
I've built two primary components for the analytics API:
1. AnalyticsProvider - provides an analytics API which can be expanded as-needed in the future
2. Page - a generic page component which views can render to capture pageview analytics. The API is very simple at this time, but it is designed to be expanded for other types of page-behaviors if needed


src/contexts/Analytics/Analytics.tsx
```diff
@@ -0,0 +1,48 @@
...
+const AnalyticsContext = createContext<AnalyticsAPI>({
+  trackPageView: () => {},
+});
+
+export const useAnalytics = () => useContext(AnalyticsContext);
+
+export const AnalyticsProvider: FunctionComponent = ({ children }) => {
+  const { pathname } = useLocation();
+
+  const prevPath = useRef<null | string>(null);
```
gabewomble, line 24: In order to prevent duplicate tracking calls, we want to store the previous location to guarantee we don't somehow fire a pageview call twice for the same url

addihorowitz (approved):

Issue #255, opened 2021-07-23
Create an error page


addihorowitz:
On failures, which might occur, we need to have an error page

gabewomble:
This issue requires two changes:

Back End:
- Serving resources from a separate domain to properly return a 404 for missing objects instead of returning a 200 with the application html (since the client is a SPA)

Front End:
- A custom error page / component when the documentation 404s

PR #260, opened 2021-07-25
fix: missing doc pages result in an unclear error 

iliapolo:
Currently, when users browse to a documentation page in a language that is either:

1. Unsupported by the package
2. Documentation isn't ready for yet
3. Documentation failed to create for

This is what we show:

Screen Shot 2021-07-25 at 11 36 11 AM

> Example: https://constructs.dev/packages/@softchef/cdk-vue/v/1.0.31?lang=python

This happens because we currently don't detect that we requested a missing document, and we simply render the `index.html` page that CloudFront serves in such scenarios. 

This PR adds this missing docs page detection by asserting that the content type of the response is the one we expect. 
It works because our doc pages have `text/markdown`, while our `index.html` has `text/html`. 

This is just a tactical stop-gap until we implement proper 404 responses for such cases on the backend. 

Following this PR, these errors will displayed as:

Screen Shot 2021-07-25 at 11 26 42 AM

The error still needs some design, but at least it's clear and coherent. 

Fixes https://github.com/cdklabs/construct-hub-webapp/issues/255
Fixes https://github.com/cdklabs/construct-hub/issues/218


src/api/package/docs.ts
```diff
@@ -19,5 +19,19 @@ export const fetchMarkdown = async (
...
+
+  // since CloudFront returns a 200 /index.html for missing documents,
+  // we assert the expected docs content type to detect these errors.
+  // TODO: switch to proper 404 responses in this case (requires backend changes)
+  const expectedContentType = "text/markdown";
+  const contentType = response.headers.get("Content-Type");
+
+  // we check 'includes' and not 'equal' because the content type contains
+  // charset as well (e.g text/markdown; charset=UTF-8)
+  if (!contentType || !contentType.includes(expectedContentType)) {
```
eladb, line 31: Would it make sense to also match something in the content of the response page (i.e. the string `"404"`)?

eladb (commented):
Is it possible to add a metric that monitors this experience? Would be good to have visibility how many times users are hitting this? (could also be something we track on the backend, but let's make sure we are tracking this)

src/views/Package/Package.tsx
```diff
@@ -18,6 +19,36 @@ interface PathParams {
...
+  if (markdown.error || assembly.error) {
+    return (
+      <Text
+        align="center"
+        fontSize="xl"
+        fontStyle="oblique"
+        wordBreak="break-word"
+      >
+        Oops, Looks like {language} documentation is not available yet. Checkout
+        other languages in the meantime!
```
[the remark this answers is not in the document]
iliapolo: @addihorowitz Do we have some support email we plug here?

src/api/package/docs.ts
iliapolo: Don't think so...what kind of situation were you envisioning? this whole assertion is strongly based on the fact we know exactly what the response should be in this case.

src/views/Package/Package.tsx
```diff
@@ -18,6 +19,36 @@ interface PathParams {
...
+      </Text>
+    );
+  }
+
+  if (markdown.loading || assembly.loading) {
+    return null;
+  }
+
+  if (!markdown.data || !assembly.data) {
```
gabewomble, line 45: Imo we can just consolidate these into one if statement

src/views/Package/Package.tsx
```diff
@@ -53,15 +84,7 @@ export const Package: FunctionComponent = () => {
...
-      {markdownResponse.data &&
-        !markdownResponse.loading &&
-        assemblyResponse.data &&
-        !assemblyResponse.loading && (
-          <PackageDocs
-            assembly={assemblyResponse.data}
-            markdown={markdownResponse.data}
-          />
-        )}
+      {docsOrError(markdownResponse, assemblyResponse, language)}
```
gabewomble, line 87: I would prefer to see this either as a function component - `<DocsOrError {...props} />` or as 
```tsx
{hasDocs ? (
  <PackageDocs {...props} />
) : (
  <ErrorMessage {...props} />
)}
```

So that it is consistent with how we define & call our function components

gabewomble (changes requested):
Just a couple small comments, feel free to dismiss this once addressed, looks good otherwise!

src/views/Package/Package.tsx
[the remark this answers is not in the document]
addihorowitz: We don't but we can create one. Will send you the email on slack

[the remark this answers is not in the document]
addihorowitz: We don't have one. We can open one. However, they can just open an issue here https://github.com/cdklabs/construct-hub-webapp/issues/new/choose. What do you think?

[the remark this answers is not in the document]
iliapolo: Rephrased and added a link to issue creation.

src/views/Package/Package.tsx
iliapolo: 👍

src/views/Package/Package.tsx
iliapolo: Done

gabewomble (approved):

PR #266, opened 2021-07-25
fix: relative images are not displayed

iliapolo:
When a README in a package repository includes a relative link to an image, i.e

```text
![](./images/access-token.png)
```

These images are not being rendered. For example:

Screen Shot 2021-07-25 at 7 27 13 PM

This happens because the constructed URL should not contain the `.git` repo suffix. Removing it will result in:

Screen Shot 2021-07-25 at 7 26 48 PM


src/components/Markdown/Markdown.tsx
```diff
@@ -110,7 +110,8 @@ export const Markdown: FunctionComponent<{
             return url;
           }
 
-          const { owner, repo } = repoConfig;
+          const owner = repoConfig.owner;
+          const repo = repoConfig.repo.replace(".git", "");
```
eladb, line 114:
```suggestion
          const repo = repoConfig.repo.replace(/\.git$/, "");
```

eladb (approved):

PR #272, opened 2021-07-26
feat: Card & Tag UX

gabewomble:
This PR does a lot of little things

New APIs:
- `createURL`: creates a string url from a base & optional search params object
- `createURLSearchParams`: creates a string search from a params object and an optional base
- `getSearchPath`: Composes `createURL` to provide an API for getting search urls
- `getPackagePath`: Composes `createURL` to provide an API for getting package profile urls
- `<PackageTag />`: A tag component which will apply a search for it's value when clicked

useCatalogSearch:
- Exposes lower level methods to allow finer control over the searchAPI instance
- Documentation around all returned values & methods

Card UX:
- Card click will link to filtered language
- Added a hover state
- Language icons link to the package with their respective language docs
- Tags updated to use `<PackageTag />`

CatalogSearch UX:
- Language dropdown now has an arrow to match other dropdowns in app

PackageHeader UX:
- Tags updated to use `<PackageTag />`

Screen Shot 2021-07-26 at 5 06 58 PM


src/components/CatalogCard/CatalogCard.tsx
```diff
@@ -59,31 +70,42 @@ export const CatalogCard: FunctionComponent<CatalogCardProps> = ({
...
+            color="blue.800"
+            data-testid={testIds.name}
+            fontWeight="semibold"
+            isTruncated
           >
-            {pkg.version}
+            {pkg.name}
+
+            <Text
+              as="span"
```
addihorowitz, line 95: Can't all these font/color/size definitions be implemented with css class?

addihorowitz (approved):
Approved. I added a comment about css class but it shouldn't postpone us from merging the code right now

src/components/CatalogCard/CatalogCard.tsx
gabewomble: We don't use plain stylesheets, we leverage a styled system from https://chakra-ui.com/docs/principles

PR #277, opened 2021-07-27
chore: nav tree ui updates

gabewomble:
Updates the Nav Tree to:
- Not cut off outlines for top & bottom elements
- Align children with their parent text rather than indent

Screen Shot 2021-07-27 at 9 56 33 AM
Screen Shot 2021-07-27 at 9 56 46 AM


addihorowitz (approved):
Approved

Issue #271, opened 2021-07-26
When you click a publisher name on the results page it should run a search for all their packages on the Construct Hub


addihorowitz:
Instead of opening the publisher page on npm, it should run a search for this publisher's packages on Construct Hub.
This behavior is relevant for the results page only.

PR #285, opened 2021-07-28
feat: search by author

iliapolo:
In the result cards, clicking on the author will now result in a search query for that author, instead of going to the author UR, which will now only be available from the operator area of the package page. 

Closes https://github.com/cdklabs/construct-hub-webapp/issues/271

### DIFF
--- a/src/components/CatalogCard/CatalogCard.tsx
+++ b/src/components/CatalogCard/CatalogCard.tsx
@@ -145,7 +145,7 @@ export const CatalogCard: FunctionComponent<CatalogCardProps> = ({
           <UILnk
             color="blue.500"
             fontSize="sm"
-            href={`./search?q=${authorName}&offset=0`}
+            href={`./search?q=${authorName}`}
             onClick={(e) => e.stopPropagation()}
           >
             {authorName}
--- a/src/components/CatalogCard/CatalogCard.tsx
+++ b/src/components/CatalogCard/CatalogCard.tsx
@@ -3,6 +3,7 @@ import {
   Divider,
   Flex,
   LinkBox,
+  Link as UILnk,
   LinkOverlay,
   Stack,
   Text,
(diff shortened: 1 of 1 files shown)


MrArnoldPalmer (approved):

gabewomble (changes requested):
Wait

src/components/CatalogCard/CatalogCard.tsx
```diff
@@ -3,6 +3,7 @@ import {
   Divider,
   Flex,
   LinkBox,
+  Link as UILnk,
```
gabewomble, line 6: typo

src/components/CatalogCard/CatalogCard.tsx
```diff
@@ -140,24 +142,14 @@ export const CatalogCard: FunctionComponent<CatalogCardProps> = ({
...
-            <Text data-testid={testIds.author} fontSize="sm" isTruncated>
-              {typeof author === "string" ? author : author.name}
-            </Text>
-          )}
+          <UILnk
+            color="blue.500"
+            fontSize="sm"
+            href={`./search?q=${authorName}`}
+            onClick={(e) => e.stopPropagation()}
+          >
```
gabewomble, line 150:
```suggestion
          <UILink
            as={Link}
            color="blue.500"
            fontSize="sm"
            to={getSearchPath({ query: authorName })}
          >
```

gabewomble: getSearchPath will need to be imported from the same file as `getPackagePath` (which is already imported)

iliapolo: Done

src/components/CatalogCard/CatalogCard.tsx
iliapolo: Danke

gabewomble (approved):

PR #287, opened 2021-07-29
fix: long api references do not load

iliapolo:
An example would be: https://constructs.dev/packages/cdk-iam-floyd/v/0.218.0

This PR puts a hard limit of 1MB on the size of the docs page, anything above that will cause only the README to be displayed, along with a notice that the API reference couldn't be rendered. 

Note that the README displayed is the transliterated README, so at least we have that added value for those packages. 

This is just a stop-gap until we are able to resolve the issue, but at least we won't crash and stall browsers. 

Screen Shot 2021-07-29 at 11 29 55 AM
Screen Shot 2021-07-29 at 12 15 53 PM

### DIFF
--- a/src/components/Markdown/Markdown.tsx
+++ b/src/components/Markdown/Markdown.tsx
@@ -13,6 +13,8 @@ import { Ul, Ol, Li } from "./List";
 import { Table, Thead, Tbody, Tfoot, Tr, Th, Td, TableCaption } from "./Table";
 import { A, Blockquote, Em, P, Pre, Sup } from "./Text";
 
+const ONE_MEGABYTE = 1024 * 1024;
+
 const components = {
   a: A,
   blockquote: Blockquote,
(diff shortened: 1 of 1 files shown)


src/components/Markdown/Markdown.tsx
```diff
@@ -115,6 +117,17 @@ export const Markdown: FunctionComponent<{
           return `https://${githubPrefix}/${owner}/${repo}/${githubSuffix}/${url}`;
         };
 
+  const byteLength = Buffer.byteLength(children);
+  if (byteLength > ONE_MEGABYTE) {
+    console.log(`Doc page is larger than 1MB, showing only README...`);
```
addihorowitz, line 122: I guess it's a debug message that we should remove?

addihorowitz (commented):
I think there's a debug console message we should remove

src/components/Markdown/Markdown.tsx
iliapolo: Removed it.

src/components/Markdown/Markdown.tsx
```diff
@@ -13,6 +13,8 @@ import { Ul, Ol, Li } from "./List";
 import { Table, Thead, Tbody, Tfoot, Tr, Th, Td, TableCaption } from "./Table";
 import { A, Blockquote, Em, P, Pre, Sup } from "./Text";
 
+const ONE_MEGABYTE = 1024 * 1024;
```
iliapolo, line 16: @gabewomble do you think thats a reasonable limit? can we make it bigger?

addihorowitz (approved):

PR #288, opened 2021-07-29
chore: rephrase "Use Construct"

iliapolo:
Something about "Use Construct" feels extremely foreign to me. 

Suggestion to change, alternatives are happily accepted.


src/views/Package/components/UseConstruct/UseConstruct.tsx
```diff
@@ -50,7 +50,7 @@ export const UseConstruct: FunctionComponent<UseConstructProps> = ({
 
   const trigger = (
     <CodePopoverTrigger disabled={!TEMP_SUPPORTED_LANGUAGES.has(language)}>
-      Use Construct
+      Install library
```
eladb, line 53:
```suggestion
      Install
```

eladb (commented):
How about just "install" (like in npm)

eladb:
(I agree that "Use Construct" is not ideal)

MrArnoldPalmer (approved):

PR #289, opened 2021-07-29
chore: disable preview banner dismissal

iliapolo:
Suggestion from @eladb to disable the ability to dismiss the preview banner - so that no one forgets. 

Screen Shot 2021-07-29 at 3 51 59 PM

### DIFF
--- a/src/components/DevPreviewBanner/DevPreviewBanner.tsx
+++ b/src/components/DevPreviewBanner/DevPreviewBanner.tsx
@@ -1,28 +1,13 @@
-import { CloseIcon, InfoOutlineIcon } from "@chakra-ui/icons";
-import {
-  Box,
-  IconButton,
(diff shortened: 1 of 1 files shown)


gabewomble:
Is this really necessary? If a user dismisses this and forgets, it's on them. It only stays closed for the duration of their session anyways (i.e until their construct hub tabs are closed)

iliapolo:
@gabewomble definitely not **really** necessary - we just floated this and figured i'd make a PR that we can agree/disagree based on what the team thinks. 

>  It only stays closed for the duration of their session anyways (i.e until their construct hub tabs are closed)

Cool, for some reason I thought its persisted across sessions. 

I'm leaning towards keeping it as is, wondering what other people think though.

iliapolo:
Team decided against this for now

PR #294, opened 2021-07-29
fix: search result card styling

MrArnoldPalmer:
Updates the spacing of items within the search results card and
customizes the Tag component to better match design.


gabewomble (approved):
🔥

PR #295, opened 2021-07-29
FAQ, need to fix links

addihorowitz:
Added content to FAQ, still need to fix the links and make we should we reference the FAQ from the menu


Chriscbr:
I thought I'd give this PR a look - I've added links and tried to polish things up a bit. I'm not totally familiar with the repo's style guide and practices, so please critique as needed. 🙂

Screenshot:

Screen Shot 2021-07-29 at 6 00 45 PM

MrArnoldPalmer (approved):
I think this is fine for now. We can replace the <p> tags and other non-themed components with <Text /> and others from @chakra at some point in the future.

PR #312, opened 2021-08-03
[POC] Incremental Static Regeneration, SSR, and SSG! (NextJs)

gabewomble:
This is a big PR! This is something of a pet project, but I've really been enjoying working on it in my free time, so I went all in on it

Summary of changes:
- Migrated all pages to nextjs
- Package Pages, Home, and Search are built with [ISR](https://vercel.com/docs/next.js/incremental-static-regeneration), meaning we get the best of SSR and SSG
- Informational pages such as FAQ are completely static

Note: a lot of diff comes from updating client-side routing logic to server side. The other big portion of diff comes from yarn.lock updates

Initial cypress suite:
Screen Shot 2021-08-04 at 4 27 34 PM


next.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  serverRuntimeConfig: {
+    apiUrl: "https://construct-hub-testing.dev-tools.aws.dev"
+  }
+}
```
gabewomble, line 5: @MrArnoldPalmer @eladb This `next.config.js` file would be a good candidate for how we can allow `construct-hub` to set server & runtime configurations on the webapp. 

We can define `serverRuntimeConfig` and `publicRuntimeConfig` properties with the following behaviors:
- `serverRuntimeConfig` is only exposed on the server
- `publicRuntimeConfig` is exposed across client / server, meaning it's perfect for concepts such as custom themes / branding

src/api/package/assembly.ts
```diff
@@ -1,7 +1,11 @@
 import * as spec from "@jsii/spec";
+import getConfig from "next/config";
 import { API_PATHS } from "../../constants/url";
 import { getAssetsPath } from "./util";
 
+const { serverRuntimeConfig } = getConfig();
+const { apiUrl } = serverRuntimeConfig;
```
gabewomble, line 7: Here's the `serverRuntimeConfig` being used for fetching server side data

next.config.js
MrArnoldPalmer: this is interesting, with the current layout though construct hub doesn't "build" the frontend, and so can't pass build time values. By the time it has access to construct-hub-webapp files, they are already built. This may be more argument for collapsing the two.

src/pages/_app.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Footer } from "components/Footer";
+import { Header } from "components/Header";
+import { ExternalLinkWarningProvider } from "contexts/ExternalLinkWarning";
+import { ShortbreadProvider } from "contexts/Shortbread";
+import { Theme } from "contexts/Theme";
+
+const App: FunctionComponent<AppProps> = ({ Component, pageProps }) => {
+  return (
+    <ShortbreadProvider>
+      <Head>
```
BryanPan342, line 16: Also since you are duplicating the <Head> calls, I would add keys to each of these meta tags! Right now we have a lot of duplicate meta tags in the rendered head tag.

See [here](https://nextjs.org/docs/api-reference/next/head)

Screen Shot 2021-08-17 at 12 08 59 PM

BryanPan342 (commented):
@gabewomble bro this looks TIGHT i like it 🎉

Very intuitive to walk through the repository and i like how you modularized everything 🦄

Got some general comments after a quick scan of the changes

src/pages/_app.tsx
```diff
@@ -0,0 +1,57 @@
...
+        />
+        {/* Meta */}
+        <meta charSet="utf-8" />
+        <meta content="width=device-width, initial-scale=1" name="viewport" />
+        <link href="/favicon.ico" rel="icon" />
+        <meta content="#ffffff" name="theme-color" />
+        <link href="/logo192.png" rel="apple-touch-icon" />
+
+        <link href="/manifest.json" rel="manifest" />
+        <script src="https://a0.awsstatic.com/s_code/js/3.0/awshome_s_code.js"></script>
```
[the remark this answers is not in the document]
gabewomble: Nice! I need to get around to fixing this as it seems to work fine, but it's definitely not ideal if NextJs is telling me to do it differently lol

src/pages/_app.tsx
gabewomble: Ooh good callout, I didn't really give too much thought into this one - I kind of assumed it behaved similar to an object spread where I could set defaults and overwrite them if needed!

BryanPan342: I believe the fix is to add a `key` property to the tags to allow them to be only generated once!

Or what I like to do is just define a `Layout` component that defines `<Head>` for me along with whatever `<Header>` and other things i need

eladb:
@gabewomble this is super cool!

BryanPan342:
Noooo

gabewomble:
To be revisited - There are technical challenges beyond converting the webapp to Next that stand in the way

PR #328, opened 2021-08-09
feat: cypress overhaul

gabewomble:
Fixes #

This work captures some changes that I've discussed with @MrArnoldPalmer to reduce the cost of maintaining tooling around the application. Here are the key changes

1. Cypress is the primary test-runner for the app
2. End to end tests (cypress) are running on builds, preventing breaking changes to the user experience
3. Unit tests will serve to test utilities, while components will be tested via cypress
4. Storybook has been removed - it doesn't really fit our application needs since we don't have our own design system. It's ultimately more overhead than we need to maintain

There are still plenty of test cases to be added to cypress, but I've added a basic suite of tests for the time being.


.projenrc.js
```diff
@@ -72,48 +69,64 @@ const project = new web.ReactTypeScriptProject({
...
+(function addJest() {
+  project.addDevDeps(
+    // "jest",
+    // "babel-jest",
+    // "ts-node",
+    "@testing-library/react",
+    "@testing-library/jest-dom",
+    "@testing-library/react-hooks",
+    "@testing-library/user-event"
+  );
```
gabewomble, line 122: This section is in anticipation of moving to next. Currently, we can't install our own version of jest as it will conflict with the jest config that react-app-rewired ships. Once that's out of the way we can uncomment this stuff

MrArnoldPalmer (approved):

PR #367, opened 2021-09-01
fix: catalog card tags zIndex too high

gabewomble:
Before:
Screen Shot 2021-08-31 at 5 18 33 PM

After:
Screen Shot 2021-08-31 at 5 18 22 PM


src/components/CatalogCard/CatalogCard.tsx
```diff
@@ -126,6 +126,7 @@ export const CatalogCard: FunctionComponent<CatalogCardProps> = ({
                     language={currentLanguage}
                     mr={1}
                     value={`"${tag}"`}
+                    zIndex="0 !important"
```
Chriscbr, line 129: Is there any way to fix this without `!important`?

Chriscbr (approved):

PR #377, opened 2021-09-08
feat: custom link config

MrArnoldPalmer:
Adds system for loading custom configuration to customize links on the
package page. This allows site owners to prioritize custom items in the
PackageDetails over the repository, license, and registry links. The
content of these links must be in the packages metadata.


src/contexts/Config/Config.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const ConfigProvider: FunctionComponent = ({ children }) => {
+  const [requestConfig, configResponse] = useRequest(fetchConfig);
+
+  useEffect(() => {
+    void requestConfig();
+  }, [requestConfig]);
+
+  return (
+    <ConfigContext.Provider value={configResponse}>
+      {children}
```
gabewomble, line 22: I'd probably suggest rendering an app-level loader here if the config response hasn't resolved

```tsx
{configResponse.data ? children : <Loader />}
```

MrArnoldPalmer: I think components that need the config should be responsible for showing loaders at lower levels if need be, I don't see the need to now show the whole app if config isn't loaded. That being said, I need to revise `null` rendering in the `PackageDetails` component.

gabewomble (approved):

PR #383, opened 2021-09-09
chore: cleanup package filter logic

gabewomble:
This update should make no changes to existing functionality. I've done this in anticipation of adding more filters / query support so that we only iterate through the catalog a single time

It also loosens up the parameters of `useCatalogResults` to only require a limit of results as language, query, and offset can all be safely defaulted


src/constants/languages.ts
```diff
@@ -36,6 +36,7 @@ export const LANGUAGE_NAME_MAP: { readonly [key in Language]: string } = {
 export const TEMP_SUPPORTED_LANGUAGES: ReadonlySet<Language> = new Set([
   Language.Python,
   Language.TypeScript,
+  Language.Java,
```
MrArnoldPalmer, line 39: We have a separate PR only for adding java support when rdy here https://github.com/cdklabs/construct-hub-webapp/pull/384/files

src/hooks/useCatalogResults/useCatalogResults.ts
```diff
@@ -15,31 +15,38 @@ export interface UseCatalogResultsOptions {
...
-      filtered = filtered.filter((item) =>
-        Object.keys(item.languages).includes(language)
-      );
-    }
+      return true;
+    });
 
-    return filtered.sort((p1, p2) => {
+    // Next sort by most recently updated. (The sort should eventually become a parameter)
+    return [...filtered].sort((p1, p2) => {
```
MrArnoldPalmer, line 49: what is the need to copy the array over before sorting?

MrArnoldPalmer (changes requested):
So this is defaulting the parameters and some refactoring? I see using the sort fn in the test but is the output of the implementation changed?

request changes just to remove the addition of java to temp supported languages just as unrelated, unless we need that for some reason I'm not thinking of?

src/hooks/useCatalogResults/useCatalogResults.ts
gabewomble: to avoid mutating the original dataset - this actually was causing the "bug" that was allowing tests to pass in the first place

src/constants/languages.ts
gabewomble: Ahh good callout, I did not mean to commit that here

MrArnoldPalmer (approved):

PR #421, opened 2021-09-23
Feat/improved search

### LINKED ISSUE #406: Refactor Search
As we look to expand our client-side search features, we should leverage a library such as [lunr](https://lunrjs.com/guides/core_concepts.html) to do the heavy lifting for us. We should implement this within our existing UI before implementing the new search UI

gabewomble:
Fixes #406 

1. Implements the `CatalogSearchAPI` which exposes a `.search()` method powered by `lunr.js`. Client side searches with Lunr behave much like traditional index based searches and will be a better experience for our end-users.
2. Exposes the new search api on the search page. Filters have not yet been setup
3. Implements the basic search results section, re-using a fair amount of code from the prior implementation

The most significant benefits we gain from using `lunr` is that text searches will no longer be nearly as rigid, and results will be sorted by relevance by default

Screen Shot 2021-09-23 at 2 14 30 PM
Screen Shot 2021-09-23 at 3 51 30 PM


src/components/PackageList/PackageList.tsx
```diff
@@ -20,21 +20,25 @@ export interface PackageListProps extends Partial<PackageListViewProps> {
   title?: string;
 }
 
-export const PackageList: FunctionComponent<PackageListProps> = ({
-  cardView = PackageCardType.Wide,
-  items,
-  loading,
-  // title,
-}) => {
-  if (loading || !(items && items.length > 0)) {
```
gabewomble, line 29: Don't want a spinner for 0 results 😅

src/api/catalog-search/util.ts
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+const getStrSort = (isAscending: boolean): SortFunction => {
+  return (p1, p2) =>
+    isAscending
+      ? p1.name.localeCompare(p2.name)
+      : p2.name.localeCompare(p1.name);
+};
+
+export const SORT_FUNCTIONS: Record<CatalogSearchSort, SortFunction> = {
```
MrArnoldPalmer, line 29: I think some basic unit tests covering asc/desc implementations would be nice here.

MrArnoldPalmer (changes requested):
I'm still looking through this but for now I think unit tests for sorting and filtering functionality would be a good addition.

gabewomble:
When testing with 5000 mock constructs, search (with filtering and sorting) execution time ranged between 30ms and 200ms, usually on the lower end

Screen Shot 2021-09-24 at 11 26 14 AM

MrArnoldPalmer (approved):

PR #445, opened 2021-09-28
feat: sort option list

### LINKED ISSUE #399: Implement search sort option
Options should / can be:

- Number of downloads asc / desc
- Name asc / desc
- Recently Updated

Secondary order for recently_updated and #downloads should always be Name asc. 

Screen Shot 2021-09-13 at 1 52 45 PM

gabewomble:
Fixes #399 

Screen Shot 2021-09-28 at 2 52 31 PM
Screen Shot 2021-09-28 at 2 52 40 PM
Screen Shot 2021-09-28 at 3 16 43 PM
Screen Shot 2021-09-28 at 3 16 48 PM


src/views/SearchRedesign/SearchResults.tsx
```diff
@@ -50,6 +54,18 @@ export const SearchResults: FunctionComponent = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [results, offset, pageLimit]);
 
+  // Reset offset and update url when filters & sort change
+  // We want to avoid doing this on first render / when a user directly navigates to a search URL
+  // so we keep a ref to prevent this
```
Chriscbr, line 59: Are all of the filters always off when a user first navigates to the search results page / is that the assumption being made here? Is this being done as an optimization or to prevent some kind of weird UI behavior?

Chriscbr (approved):
LGTM, just had one non-blocking question below.

src/views/SearchRedesign/SearchResults.tsx
gabewomble: So the main functionality of this side effect is to effectively reset the search when any of the filters or sort change. I.E, if I'm on the 4th page of results and change my sort or filter by a different language set, I should go back to the first page.

The check on the first render is to prevent the scenario where a user navigates directly to a url such as `/search?q=foo&offset=3` and the offset would be reset. I don't imagine users sharing search result URLs as a common scenario, however, we want to avoid breaking that experience

PR #446, opened 2021-09-28
feat: display custom package tags

### LINKED ISSUE #340: Presentational Tag Support
The current "tags" displayed on search result cards come from the list of keywords defined in the package.json of a package. However we want to allow displaying additional information such as whether the library is maintained by AWS or another trusted vendor. This could be connoted by an "official" tag and the conditions for what packages have this displayed should be configurable by the owners of the construct-hub instance.

This can be achieved some build time configuration allowing users to pass in an object describing a tag and its appearance, alongside a predicate function that dictates to what packages the tag is applied.

example `ch.config.ts`
```ts
export const config = {
  // order denotes priority of display, IE 'default' displays first if applied.
  tags: [{
    name: 'default',
    color: 'green',
    // predicate accepts package metadata and returns a boolean.
    predicate: (meta) => {
      return meta.author === 'AWS'
    }
  }]
}
```
This solution allows the frontend to handle displaying all of these "presentational" tags since this is essentially just computed state and there is nowhere to store this on the backend currently. Additionally, these tags must be searchable. Since all of the search logic is handled on the frontend currently, all we have to do is run every package in the catalog through all provided predicate functions, and add new keywords to the package list and use that new computed catalog to search against. If search logic is ever moved to the backend, or somewhere other than the react app, this will require computing and storing these somewhere that the search logic has access to.

MrArnoldPalmer:
Displays custom package tags on the package page. Each package tag with
a unique color creates a new variant in the chakra theme.

Fixes #340

### DIFF
--- a/src/contexts/Theme.tsx
+++ b/src/contexts/Theme.tsx
@@ -1,10 +1,15 @@
 import { ChakraProvider } from "@chakra-ui/react";
 import type { FunctionComponent } from "react";
-import { theme } from "../theme";
+import { PageLoader } from "../components/PageLoader";
+import { makeTheme } from "../theme";
+import { useConfig } from "./Config";
 
 export const Theme: FunctionComponent = ({ children }) => {
-  return (
-    <ChakraProvider resetCSS theme={theme}>
+  const { loading, data } = useConfig();
+  return loading ? (
+    <PageLoader />
+  ) : (
+    <ChakraProvider resetCSS theme={makeTheme(data!)}>
       {children}
     </ChakraProvider>
   );
--- a/src/api/config/index.ts
+++ b/src/api/config/index.ts
@@ -1,9 +1,14 @@
 import { API_PATHS } from "../../constants/url";
 
-export interface PackageLinksConfig {
-  name: string;
-  value: string;
-  displayText?: string;
+export interface PackageLinkConfig {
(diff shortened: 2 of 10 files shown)


src/contexts/Theme.tsx
```diff
@@ -1,10 +1,15 @@
...
+import { PageLoader } from "../components/PageLoader";
+import { makeTheme } from "../theme";
+import { useConfig } from "./Config";
 
 export const Theme: FunctionComponent = ({ children }) => {
-  return (
-    <ChakraProvider resetCSS theme={theme}>
+  const { loading, data } = useConfig();
+  return loading ? (
+    <PageLoader />
```
gabewomble, line 10: IMO, we should move the UI providers such as `Theme`, `ExternalLinkWarning`, and `CardView` (to be deleted) into `App.tsx` where we could just do a component level `if (configLoading) return <PageLoader />` so we can avoid importing PageLoader into multiple files

gabewomble: Or actually on second thought, we could just allow `makeTheme`'s data parameter to be optional and let it update when `data` updates since it's all hooked into react context.

gabewomble (commented):
First pass looks great! Had some minor feedback on `Theme.tsx`. Going to test this locally as well

PR #448, opened 2021-09-29
feat: consume featured package configuration

### LINKED ISSUE #439: Consume Home Config from BE
Consume the Home Page config defined by https://github.com/cdklabs/construct-hub/pull/356

Chriscbr:
Fixes #439

### DIFF
--- a/src/views/HomeRedesign/HomeRedesign.tsx
+++ b/src/views/HomeRedesign/HomeRedesign.tsx
@@ -1,15 +1,28 @@
-import { Flex, Grid, Heading } from "@chakra-ui/react";
-import type { FunctionComponent } from "react";
-import { PackageList } from "../../components/PackageList";
+import { Flex, Grid } from "@chakra-ui/react";
+import { FunctionComponent, useEffect, useState } from "react";
+import { DEFAULT_FEATURED_PACKAGES } from "../../api/config";
 import { Page } from "../../components/Page";
-import { useCatalogResults } from "../../hooks/useCatalogResults";
+import { useConfigValue } from "../../hooks/useConfigValue";
 import { Hero } from "./Hero";
+import { HomeSection, HomeSectionProps } from "./HomeSection";
 import { InfoPanel } from "./InfoPanel";
 
 export const HomeRedesign: FunctionComponent = () => {
-  const { page } = useCatalogResults({
-    limit: 10,
-  });
+  const homePackages = useConfigValue("featuredPackages");
+
+  const [sectionData, setSectionData] = useState<HomeSectionProps[]>([]);
+
+  useEffect(() => {
+    let config = homePackages;
+    if (!config) {
+      config = DEFAULT_FEATURED_PACKAGES;
+    }
+    setSectionData(config.sections);
+  }, [homePackages]);
+
+  const sections = sectionData.map((section) => (
+    <HomeSection key={section.name} {...section}></HomeSection>
+  ));
 
   return (
     <Page
@@ -31,13 +44,7 @@ export const HomeRedesign: FunctionComponent = () => {
           templateColumns="2fr 1fr"
           templateRows="1fr"
         >
-          <Flex direction="column">
-            <Heading as="h3" color="blue.800" mb={3} size="md">
-              Recently Updated
-            </Heading>
-
-            <PackageList items={page} />
-          </Flex>
+          <Flex direction="column">{sections}</Flex>
           <InfoPanel />
         </Grid>
       </Flex>
--- a/src/views/HomeRedesign/HomeSection.tsx
+++ b/src/views/HomeRedesign/HomeSection.tsx
@@ -1,0 +1,68 @@
+import { Heading } from "@chakra-ui/react";
+import { Fragment, FunctionComponent, useMemo } from "react";
+import { CatalogPackage } from "../../api/package/packages";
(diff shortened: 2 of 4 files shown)


Chriscbr:
Adding the do-not-merge label since this depends on changes in #356 (but is still ready to be reviewed).

src/views/HomeRedesign/HomeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+            return accum;
+          }
+        },
+        {}
+      );
+      packages = showPackages.map((p) => filteredMap[p]);
+    }
+
+    return packages;
+  }, [data?.packages, error, loading, showLastUpdated, showPackages]);
```
gabewomble, line 57: This is my own opinion, but I believe this memo should only return packages if either `showLastUpdated` or `showPackages` is defined - as otherwise it means we received an invalid config. As such, we should default `packages` to undefined rather than `data.packages`. We would then want to have our component conditionally return null if `results` is not defined like the following:

```tsx
if (!results) {
  console.warn("Helpful message about your invalid config");
  return null;
}

return (
  ...
);
```

src/views/HomeRedesign/HomeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const results = useMemo(() => {
+    if (loading || error || !data?.packages) return [];
+
+    let packages = data.packages;
+
+    if (showLastUpdated) {
+      // sort packages by date
+      packages = [...packages]
+        .sort((p1, p2) => {
+          const d1 = new Date(p1.metadata.date);
```
gabewomble, line 29: Could we use the `SORT_FUNCTIONS[CatalogSearchSort.PublishDateDesc]` helper that `CatalogSearchAPI` uses in place of this? I realize the code is identical, but it's an easy opportunity to be a little DRYer

src/views/HomeRedesign/HomeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+          if (d1 === d2) {
+            return 0;
+          }
+          return d1 < d2 ? 1 : -1;
+        })
+        .slice(0, showLastUpdated);
+    } else if (showPackages) {
+      // find the specific packages from the catalog and
+      // return them in the same order
+      const filteredMap = [...packages].reduce(
```
gabewomble, line 40: micro nitpick, but there's no need to copy `packages` here since `.reduce` doesn't mutate the original

src/views/HomeRedesign/HomeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+              [p.name]: p,
+              ...accum,
+            };
+          } else {
+            return accum;
+          }
+        },
+        {}
+      );
+      packages = showPackages.map((p) => filteredMap[p]);
```
gabewomble, line 53: There's an un-exported helper in `views/PackageLatest` which finds the latest version of a package by name. It may make sense to make that a generic util, and have something like the following:
```ts
packages = showPackages.map(findPackage);
```

src/views/HomeRedesign/HomeSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    return packages;
+  }, [data?.packages, error, loading, showLastUpdated, showPackages]);
+
+  return (
+    <Fragment key={name}>
```
gabewomble, line 60: React quirk, but there's actually no need to pass a key here since it isn't within a loop

src/views/HomeRedesign/HomeRedesign.tsx
```diff
@@ -1,15 +1,28 @@
...
+    let config = homePackages;
+    if (!config) {
+      config = DEFAULT_FEATURED_PACKAGES;
+    }
+    setSectionData(config.sections);
+  }, [homePackages]);
+
+  const sections = sectionData.map((section) => (
+    <HomeSection key={section.name} {...section}></HomeSection>
+  ));
```
gabewomble, line 25: We could likely make this component stateless and just do something like

```tsx
const homePackages = useConfigValue("featuredPackages");
const sections = (homePackages ?? DEFAULT_FEATURED_PACKAGES).sections;

return (
  ...
  {sections.map((section) => <HomeSection key={section.name} {...section} />)}
  ...
);
```

src/views/HomeRedesign/HomeSection.tsx
Chriscbr: Good point, if the code doesn't break/return it might try rendering all of the packages in the catalog which would be pretty bad...

src/views/HomeRedesign/HomeSection.tsx
Chriscbr: Ah makes sense, TIL.

src/views/HomeRedesign/HomeSection.tsx
gabewomble: I didn't even think of that edge-case, but yeah that'd cause a pretty noticeable stutter 😅

src/views/HomeRedesign/HomeSection.tsx
Chriscbr: I wasn't sure what the best place to put the utility function was so I put it in `api/package/util.ts`.

gabewomble (approved):

PR #450, opened 2021-09-29
fix: show loading spinner when switching languages

Chriscbr:
Improves the user experience by showing a loading spinner when you are switching between languages (for some docs, this can take multiple seconds). The existing experience is that a loading spinner only appears when you load the page for the first time.

Screenshot:
Screen Shot 2021-09-29 at 2 43 18 PM


gabewomble (approved):
Nice!

PR #454, opened 2021-10-01
feat: CDKType + version filter

gabewomble:
Fixes #

This PR implements the cdk type filter as well as cdk version filter if applicable. A CDK Version filter can only be selected if a CDK Type is first selected, and there is more than one major version available to select from

I've also added hints to the filters to help users learn how they behave

Screen Shot 2021-09-30 at 3 26 29 PM
Screen Shot 2021-09-30 at 3 26 34 PM
Screen Shot 2021-09-30 at 3 29 07 PM
Screen Shot 2021-09-30 at 3 29 12 PM

Note: Writing unit tests, but the rest is ready for review


src/views/SearchRedesign/Filter.tsx
```diff
@@ -23,9 +40,34 @@ export const Filter: FunctionComponent<FilterProps> = ({
...
+            <PopoverContent>
+              <PopoverArrow />
+              <PopoverCloseButton />
+              <PopoverHeader>Hint: {name}</PopoverHeader>
+              <PopoverBody>
+                <Text>{hint}</Text>
+              </PopoverBody>
+            </PopoverContent>
+          </Popover>
+        ) : null}
```
Chriscbr, line 69: I wonder if we should use the same tooltip component/UI across different parts of the website? I noticed they look a little different between screenshots in different PRs. Also the tooltip text here is a larger font size than the text on the sidebar - maybe we could try making the text smaller?

Chriscbr (changes requested):
Looks great, just added a few small comments!

Chriscbr (approved):

PR #462, opened 2021-10-05
remove duplicate text in FAQ

TheRyanBurke:
the phrase "which is a TypeScript-based" appeared twice in a row


Chriscbr (approved):
Thank you @TheRyanBurke!

PR #466, opened 2021-10-05
feat: display cdktype icon on card

gabewomble:
Fixes https://github.com/cdklabs/construct-hub/issues/303

Displays the CDK Type Icon for package cards that have an associated type. Hovering these icons will display a tooltip with the cdk type's name and optionally a version (if defined)

Screen Shot 2021-10-05 at 12 22 40 PM
Screen Shot 2021-10-05 at 12 22 55 PM
Screen Shot 2021-10-05 at 12 23 12 PM


gabewomble:
> Looks awesome. 😄 Is it possible to get the CDKTF / AWS CDK icons with transparency?

I don't have access to any (yet), but I'll see what I can find

Chriscbr (approved):

PR #475, opened 2021-10-07
feat: new search bar

### LINKED ISSUE #408: Update Header Search to be an input
In our new designs, the search in the header will be a standalone input, rather than a modal trigger. This will lose the functionality to apply search filters from non search pages, but these can still be applied on the search page and should be a simpler experience overall

### LINKED ISSUE #440: Support Search from New Home Page
Integrate the new search functionality with the new home page UI. Currently, this UI is non-functional

gabewomble:
Fixes #408 
Fixes #440 

Implements a SearchBar with an optional overlay on focus and search suggestions.

This component is used on the existing Header, new Home Page, and new Search Page

Screen Shot 2021-10-08 at 9 55 09 AM
Screen Shot 2021-10-08 at 9 55 26 AM
Screen Shot 2021-10-08 at 9 55 44 AM


src/api/catalog-search/catalog-search.ts
```diff
@@ -227,7 +227,10 @@ export class CatalogSearchAPI {
         if (frameworkName) {
           const entry = frameworks[frameworkName];
 
-          if (majorVersion && !entry.majorVersions.includes(majorVersion)) {
+          if (
+            majorVersion !== undefined &&
```
MrArnoldPalmer, line 231: why change to `!== undefined`?

src/components/Header/HeaderSearch.tsx
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Renders a SearchBar at desktop resolutions and shows a search icon which opens a modal on mobile resolutions
+ */
+export const HeaderSearch: FunctionComponent = () => {
+  const searchModal = useDisclosure();
+  const { pathname } = useLocation();
+
+  if ([ROUTES.HOME, ROUTES.SEARCH].some((path) => path === pathname)) {
+    return null;
+  }
```
MrArnoldPalmer, line 19: Can't we just use react-router `Route` here normally to not render on home/search?

```jsx
  <Switch>
    <Route path={[ROUTES.HOME, ROUTES.Search]} />
    <Route path="*">
      <>{mySuperComponent}</>
    </Route>
  </Switch>
```

gabewomble: Yeah we totally could, this is just the same code from the former `SearchButton.tsx` did. But I do think the react-router suggestion is a good one

PR #479, opened 2021-10-08
chore: e2e xss tests

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-internal/issues/35

This PR implements a test suite of xss tests for search inputs and urls across the site. Let me know if you can think of any missing cases, so far I've got the following cases

Home (Redesign + Existing)
- Inputs do not execute JS or HTML

Search (Redesign + Existing)
- Inputs do not execute JS or HTML
- URL query params do not execute JS or HTML

Package Page
- Markdown will not execute JS or HTML


cypress/integration/xss.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+      testInput({
+        url: "/search",
+        input: alertJS,
+      });
+    });
+  });
+});
+
+describe("XSS - GA Featureset", () => {
+  // describe("Home Page - Input", () => {
```
gabewomble, line 142: These tests are dependent on the new SearchBar PR to be merged first

gabewomble:
Looks like all of these failed builds boil down to the markdown test failing in CI, while it passes locally. Unfortunately it's not a matter of timeouts, but rather that the response we get back in CI doesn't give us markdown. Going to see what's up monday

You can watch the test runs yourself by downloading the video artifacts for a commit's workflow (click the red X to view the page)

Screen Shot 2021-10-08 at 4 13 04 PM

PR #497, opened 2021-10-18
feat: add download counts to package cards

### LINKED ISSUE #463: Display number of downloads on package cards
Requires parts of https://github.com/cdklabs/construct-hub/issues/290

Chriscbr:
Fixes #463

### DIFF
--- a/src/components/PackageCard/Details.tsx
+++ b/src/components/PackageCard/Details.tsx
@@ -1,5 +1,7 @@
+import { DownloadIcon } from "@chakra-ui/icons";
 import { Text } from "@chakra-ui/react";
-import type { FunctionComponent, ReactChild } from "react";
+import { FunctionComponent, ReactChild, useMemo } from "react";
+import { useStats } from "../../contexts/Stats";
 import { useLanguage } from "../../hooks/useLanguage";
 import { getSearchPath } from "../../util/url";
 import { NavLink } from "../NavLink";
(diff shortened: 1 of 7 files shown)


src/components/PackageCard/Details.tsx
```diff
@@ -30,12 +36,26 @@ export const Details: FunctionComponent = () => {
     author,
     metadata: { date },
     version,
+    name,
   } = usePackageCard();
 
+  const { data } = useStats();
+  const downloads: number | undefined = useMemo(() => {
+    return data?.packages?.[name]?.downloads?.npm;
+  }, [name, data]);
```
gabewomble, line 45: Nitpick, but I think you can drop the useMemo since we're only doing property access for a number value here. A reason we would potentially want to memoize this value is if we were accessing an object (or array) and needed to prevent it from causing unwanted renders due to react's shallow object equality checks i.e `object1 === object2`

Chriscbr: Makes sense! 👍

src/components/PackageCard/Details.tsx
```diff
@@ -30,13 +36,26 @@ export const Details: FunctionComponent = () => {
...
 
+  const { data } = useStats();
+  const downloads: number | undefined = data?.packages?.[name]?.downloads?.npm;
+
   const authorName = typeof author === "string" ? author : author.name;
 
   return (
     <>
-      <Detail data-testid={testIds.version} label="Version" value={version} />
+      {downloads !== undefined && downloads >= 10 ? (
```
Chriscbr, line 49: BTW - just wanted to bring attention to this. I suggest here not showing downloads if the number of downloads is very low, since according to NPM, there can be a bit of noise in the signal due to downloads just coming from NPM mirrors, which does not represent actual usage of the library. Open to removing this, though.

PR #501, opened 2021-10-19
feat: upgrade construct-hub whenever a release is made

Chriscbr:
Resolves part of https://github.com/cdklabs/construct-hub-internal/issues/56

This is a partial solution to improving deployment frequency. https://github.com/cdklabs/construct-hub by default only picks up changes from the web app once a day, but now it will pick up the new changes every time a new release is made.

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -234,6 +234,26 @@ rewireCRA(buildTask);
 rewireCRA(project.tasks.tryFind("test"));
 rewireCRA(project.tasks.tryFind("dev"));
 
+// trigger construct-hub to pick up changes from construct-hub-webapp
+// whenever a new release is made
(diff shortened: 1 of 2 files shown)


.projenrc.js
```diff
@@ -234,6 +235,28 @@ rewireCRA(buildTask);
 rewireCRA(project.tasks.tryFind("test"));
 rewireCRA(project.tasks.tryFind("dev"));
 
+// trigger construct-hub to pick up changes from construct-hub-webapp
+// whenever a new release is made
```
eladb, line 239: Let's add this as a projen feature!

```ts
project.release.triggerUpgrade('cdklabs/construct-hub', {
  githubToken: ...
})
```

Issue #427, opened 2021-09-27
Package Page: Display AWS CDK's version with which the package was created


addihorowitz:
Customers who are not working with AWS CDK V2 yet needs this information in order to see that their construct is compatible with the displayed package

Issue #330, opened 2021-08-11
Construct page header takes too much space


alexpulver:
When I look for information about a construct, I am less interested in metadata. The metadata could be a single line perhaps. Then more of the documentation would be visible without scrolling

![image](https://user-images.githubusercontent.com/4362270/128972258-a3ca527f-8526-4713-bb70-b61ef8616e69.png)

gabewomble:
I think our sentiment is that the current design of the heading area in general needs some love. It indeed takes up a fair amount of space, and doesn't yet have enough information for the space to be properly used. We're currently revisiting the usability of our UX & UI and this will be one of the areas of focus

PR #547, opened 2021-10-28
 feat: New Package Page Header

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-webapp/issues/427
Fixes https://github.com/cdklabs/construct-hub-webapp/issues/330

This work ended up being larger than desired, with lots of little fixes being implemented along the way. I'll try to summarize all the minor changes below, but the primary work done here is implementing the new package heading and reorganizing package page code

Screen Shot 2021-10-28 at 3 19 30 PM

Screen Shot 2021-10-28 at 3 40 11 PM


Chriscbr:
> Also PR title should be feat: xxxx not sure why this passes our mergify checks.

I think this kind of check is only enabled in the AWS CDK repo right now - but this is a great idea for a projen feature! 😄

MrArnoldPalmer (approved):

PR #558, opened 2021-11-01
feat: enable dotnet language docs

MrArnoldPalmer:
Adds dotnet (csharp) to supported language list. Adds a new data
structure mapping language name to generated language filename, since
`dotnet` is the supported target and `csharp` is the language we
generate docs for.


Chriscbr (changes requested):
What does the install instructions for .NET packages look like? (not sure if this was added in a previous PR already)

.projenrc.js
```diff
@@ -192,7 +192,10 @@ project.npmignore.addPatterns("src/__fixtures__");
 project.gitignore.exclude("public/data");
 
 // Proxy requests to awscdk.io for local testing
-project.package.addField("proxy", "https://constructs.dev/");
+project.package.addField(
+  "proxy",
+  "https://construct-hub-testing.dev-tools.aws.dev/"
+);
```
Chriscbr, line 198: Remove?

Chriscbr (approved):

PR #574, opened 2021-11-03
chore: cleanup unused exports

gabewomble:
This is a dev tooling update which adds a script to analyze all unused exports and print the output in a readable format. Using this tool, I was able to remove a fair amount of dead code and unused exports

Before:

Screen Shot 2021-11-03 at 3 07 59 PM

After:

Screen Shot 2021-11-03 at 3 17 11 PM


.projenrc.js
```diff
@@ -170,6 +170,12 @@ const project = new web.ReactTypeScriptProject({
   project.eslint.addIgnorePattern("jest.config.ts");
 })();
 
+// This task is used to analyze dead code
+(function addAnalyzeExports() {
+  project.addDevDeps("ts-unused-exports");
+  project.addTask("analyze-exports", { exec: "node scripts/analyze-exports" });
```
Chriscbr, line 176: nit: We should add `description`'s for tasks

Chriscbr (approved):
Neat tool! 😄

PR #610, opened 2021-11-05
chore: shadow-ban packages with a specific keyword

RomainMuller:
We have a probe package that gets a new version published every 3 hours,
and we don't want this to pollute search results, or steal a "recently
updated" slot away from a more deserving package...

We are hence registering the `construct-hub/hide-from-search` keyword
on such packages, which should effectively cause the package to never
be shown on search results, or in the "what's new" section.

### DIFF
--- a/src/api/catalog-search/catalog-search.ts
+++ b/src/api/catalog-search/catalog-search.ts
@@ -65,20 +65,26 @@ export class CatalogSearchAPI {
   public readonly constructFrameworks: CatalogConstructFrameworks;
 
   constructor(catalogData: CatalogPackage[], stats: PackageStats) {
-    const catalogMap = catalogData.reduce((map, pkg) => {
-      const { name, version } = pkg;
-      const id = [name, version].join("@");
+    const catalogMap = catalogData
+      // Packages with the "construct-hub/hide-from-search" keyword are shadow-banned from search results
+      .filter((pkg) => !pkg.keywords?.includes('construct-hub/hide-from-search'))
+      .reduce(
+        (map, pkg) => {
+          const { name, version } = pkg;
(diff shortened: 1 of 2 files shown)


src/api/catalog-search/catalog-search.ts
```diff
@@ -65,20 +65,25 @@ export class CatalogSearchAPI {
...
+        const downloads = stats.packages[name]?.downloads?.npm ?? 0;
+
+        map.set(id, {
+          ...pkg,
+          downloads,
+          id,
+        });
+
+        return map;
+      }, new Map<string, any>());
```
gabewomble, line 86: just two nits here:
- we could consolidate the `.filter` and `.reduce` by only calling `map.set` based on the filter condition. I dont' have any preference here
- instead of `new Map<string, any>()` perhaps we could use something like `new Map<string, ExtendedCatalogPackage>()`?

gabewomble (approved):

gabewomble:
Looks good to me! Added do-not-merge to give you a chance to (optionally) respond to comments before merging

src/api/catalog-search/catalog-search.ts
RomainMuller: > * we could consolidate the .filter and .reduce by only calling map.set based on the filter condition. I dont' have any preference here

Yeah I considered this, but I actually like the filter to be a separate stage in the pipeline. It makes it less likely to mis-interpret... The explaining comment feels a lot more local this way.

> instead of new Map<string, any>() perhaps we could use something like new Map<string, ExtendedCatalogPackage>()?

I had considered this. The previous syntax implicitly defaulted to `any` here, so I kept that.

Issue #560, opened 2021-11-02
Separate types into pages


MrArnoldPalmer:
In order to improve rendering performance, we are changing the way package pages are rendered. In order to avoid large portions of blocking CPU usage rendering the api reference, we will break each type into its own page similar to the current cdk reference documentation. This means the readme will show as a single page, and each construct, enum, class, and struct will render as their own page. The left nav will continue to display as is allowing users to see the overall structure of a given package.

PR #611, opened 2021-11-05
feat: break up api rendering

MrArnoldPalmer:
Break up the rendering of a packages documentation into multiple pages.
The root package page now renders the readme, and each type within the
packages API reference is rendered on its own page. Items within the
readme and types are still linkable via hash links that are scrolled to
by the PackageDocs component.

The rendering is broken up by splitting the incoming markdown on the
`API Reference` header and then on subsequent headers within the api
reference markdown string. These are then parsed into menu items and a
hash map of `typeid -> type markdown content`.

Adds unit tests for markdown parsing logic to ensure outputs remain
stable.

Removes 1MB size limit for rendering markdown documents.

NOTE: [`setext`](https://github.github.com/gfm/#setext-headings) style md
headings are not currently supported and will be added in a follow up. See
https://github.com/cdklabs/construct-hub-webapp/issues/629

fix: #560


src/views/Package/PackageState.tsx
```diff
@@ -104,6 +283,37 @@ export const PackageStateProvider: FunctionComponent = ({ children }) => {
...
+          if (isMounted) {
+            setProcessMd(String(file));
+          }
+        });
+    }
+
+    return () => {
+      isMounted = false;
+    };
+  }, [markdownResponse]);
```
Chriscbr, line 304: I did some profiling on some of the larger docs like `@aws-cdk/aws-ec2` that still take around 20-30 seconds to load, and it seems like this useEffect call is the main bottleneck. If I remove this (and instead pass `markdownResponse.data` to `parseMarkdownStructure` below), then the docs load fast as expected. Let's see if there's a way to defer processing emoji instead of doing it up front

src/views/Package/PackageDocs.tsx
```diff
@@ -124,7 +77,14 @@ export const PackageDocs: FunctionComponent<PackageDocsProps> = ({
...
         }}
       >
-        {markdown}
+        <Switch>
+          <Route exact path={path}>
+            <PackageReadme />
+          </Route>
+          <Route exact path={`${path}/api/:typeId`}>
+            <PackageTypeDocs />
+          </Route>
```
Chriscbr, line 86: Would using `LazyRoute` here improve performance at all?

src/views/Package/PackageState.tsx
MrArnoldPalmer: yeah adding emoji support was done last minute once I noticed it breaking on construct-hub. Wasn't expecting a remark parse with a single plugin to be this slow, but I suspect it is `remark-emoji` itself and stacking `Array.find`s for each emoji encountered. https://github.com/rhysd/remark-emoji/blob/master/index.js#L21 Will look into handling the replacement ourselves.

MrArnoldPalmer: Fixed by manually handling emoji rendering in menu items from header strings using `node-emoji`.

src/views/Package/PackageDocs.tsx
MrArnoldPalmer: Very little, basically saves including PackageTypeDocs and/or PackageReadme components if youre on one or the other, which are both pretty tiny.

eladb (commented):
Yeah!

Curious- did you have a chance to benchmark the load time for large modules against the previous version?

addihorowitz:
> Yeah!
> 
> Curious- did you have a chance to benchmark the load time for large modules against the previous version?

@MrArnoldPalmer It should be very easy to adapt the performance test script to test and compare the performance before and after.

MrArnoldPalmer:
> Yeah!
> 
> Curious- did you have a chance to benchmark the load time for large modules against the previous version?

Only with local builds, which is not a super accurate number because of dev assets instead of our optimized product assets. Though the difference was clear and that should stand. 

I can run some comparisons in my account and report back.

MrArnoldPalmer:
@eladb @addihorowitz here are numbers comparing the ec2 package rendering performance before and after this PR.

Root documentation page rendering time with loading assets:
New: approx 1.5 seconds
ec2optimized
Old: approx 14.5 seconds
ec2old

These screenshots are the browser tools timeline showing the first frames that include rendered markdown content of the readme.

Here are also the lighthouse performance reports that you can view in your browsers dev tools for more detailed information. You can see that though performance is obviously greatly improved, the performance score in fact goes DOWN. This is (as far as we know) because the old version of the site takes so long to render that lighthouse and pagespeedinsights actually assume that it is done much earlier than in reality, and show first contentful paint around 2 seconds, and time to interactive on desktop around 5, though we know this isn't true.

[ec2reports.zip](https://github.com/cdklabs/construct-hub-webapp/files/7499091/ec2reports.zip)

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+    }
+  });
+
+  return {
+    readme: segments.join(separator),
+    apiReference: types,
+    menuItems,
+  };
+};
+
```
gabewomble, line 237: I'm not sure this is necessarily page-level state or needs to be. Do you think it would make sense to encapsulate this logic within a `PackageDocsState.tsx` or even just a util that is used by the docs component?

MrArnoldPalmer: idk I think `menuItems` + `apiReference` + `readme` is basically page level for the package page. It's not state, it's all derived, but if the md doc itself is page level I'd argue that all of this is too. We could definitely compute this within the component but thinking forward, if this is the format we ideally would want the backend to send to us, I think doing this type of normalization right after receiving the response is the correct place.

In practice  we can put it wherever, but creating another context seems unnecessarily contrived when all of this processing is done from the result of the md document and the document itself is basically no longer useful. I'd just say make a utils file right next to this that exports `parseMarkdownStructure` but that's all I'd want to separate off the top of my head.

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+  items: MenuItem[],
+  item: MenuItem
+): MenuItem[] => {
+  const last = items[items.length - 1];
+
+  if (last && last.level < item.level) {
+    last.children = appendMenuItem(last.children, item);
+    return items;
+  }
+  return [...items, item];
```
Chriscbr, line 89: nit: Might be slightly faster to just call `push` here

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+};
+
+export const splitOnHeaders = (md: string, level: number): string[] => {
+  if (!md) {
+    return [];
+  }
+
+  const regex = new RegExp(`(^#{1,${level}}[^#].*)`, "gm");
```
Chriscbr, line 97: What happens if the author uses [setext headings](https://github.github.com/gfm/#setext-headings) to delineate their headers? Does the markdown still rendered (even if the heading is not identified)? Perhaps this could be added as a test case?

It's probably OK if this isn't supported since it only affects user README's, but IMO we should call this out in a comment and in the PR description.

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+  return (
+    md
+      ?.split(regex)
```
Chriscbr, line 100: ? should not be needed here

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+ * in the api reference.
+ */
+export const parseMarkdownStructure = (
+  input: string,
+  {
+    scope,
+    language,
+    name,
+    version,
+  }: { scope?: string; language: string; name: string; version: string }
```
Chriscbr, line 150: If possible
```suggestion
  }: { scope?: string; language: Language; name: string; version: string }
```

src/views/Package/PackageTypeDocs.tsx
```diff
@@ -0,0 +1,42 @@
...
+const usePackageTypeDocs = () => {
+  const { typeId }: { typeId?: string } = useParams();
+  const { apiReference } = usePackageState();
+
+  const content = apiReference?.[typeId ?? ""];
+  if (!typeId || !content) {
+    return;
+  }
+
+  return content;
```
Chriscbr, line 17: nit: Readability (I think this is equivalent, but could be wrong)
```suggestion
  if (typeId) {
    return apiReference?.[typeId];
  }

  return;
```

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+  const last = items[items.length - 1];
+
+  if (last && last.level < item.level) {
+    last.children = appendMenuItem(last.children, item);
+    return items;
+  }
+  return [...items, item];
+};
+
+export const splitOnHeaders = (md: string, level: number): string[] => {
```
Chriscbr, line 92:
```suggestion
export const splitOnHeaders = (md: string, maxLevel: number): string[] => {
```

src/views/Package/PackageState.tsx
```diff
@@ -50,6 +76,165 @@ export const usePackageState = () => {
...
+    language,
+    name,
+    version,
+  }: { scope?: string; language: string; name: string; version: string }
+): { readme: string; apiReference: Types; menuItems: MenuItem[] } => {
+  const nameSegment = scope ? `${scope}/${name}` : `${name}`;
+  const basePath = `/packages/${nameSegment}/v/${version}`;
+  const langQuery = `?${QUERY_PARAMS.LANGUAGE}=${language}`;
+  const separator =
+    '\n# API Reference <span data-heading-title="API Reference" data-heading-id="api-reference"></span>\n';
```
Chriscbr, line 156: If the package's README is empty or non-existent, will there still be a newline at the beginning? If not, this separator might be found zero times (even though there is a "API Reference" heading in the file), and I think there's a possibility we could end up with a bug where "API Reference" appears twice in the table of contents.

src/views/Package/PackageState.tsx
MrArnoldPalmer: yeah after fiddling with this briefly I think it's best handled in a follow up cause it will take some time to get the regexes correct and change the data flow slightly. Adding a note to the function doc string and updating the PR description.

src/views/Package/PackageState.tsx
MrArnoldPalmer: Extracted into a `util.ts` file for now. Down to revisit restructuring contexts/hooks here in a follow up if you want to discuss it more.

src/views/Package/PackageState.tsx
MrArnoldPalmer: removed the leading `\n`. I believe we can rely on it's presence as much as we can rely on any of the structure within the document returned by jsii-docgen (which in this functionality is obviously a lot), as we join all segments of the larger md document with `\n` and the readme is always it's own segment. So at minimum we should have `readmetext\napireference` if not `readmetext\n\napireference`. Regardless we don't need it and its one less assumption so 👍🏻

Chriscbr (approved):

PR #623, opened 2021-11-09
fix: replace service worker with no-op worker

Chriscbr:
In #571 we updated the default JavaScript code to try and unregister service workers. However, the same service-worker.js file is still being served at the root of the website, which may cause browsers that still have the old service worker installed to think there has been no changes to the service worker, so it may continue serving old cached content. A more proper fix as described by [this Stackoverflow post](https://stackoverflow.com/questions/33986976/how-can-i-remove-a-buggy-service-worker-or-implement-a-kill-switch/38980776) is to replace the service worker with a no-op worker that does not try serving any cached content.

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -264,6 +264,12 @@ project.release.addJobs({
   },
 });
 
+// replace default service worker script with no-op worker
+const replaceWorker = project.addTask("replace-worker");
+replaceWorker.exec("cp src/no-op-sw.js build/service-worker.js");
+replaceWorker.exec("rm build/service-worker.js.map");
+project.compileTask.spawn(replaceWorker);
+
 project.synth();
 
 /**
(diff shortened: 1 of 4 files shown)


.projenrc.js
```diff
@@ -264,6 +264,12 @@ project.release.addJobs({
   },
 });
 
+// replace default service worker script with no-op worker
+const replaceWorker = project.addTask("replace-worker");
+replaceWorker.exec("cp src/no-op-sw.js build/service-worker.js");
+replaceWorker.exec("rm build/service-worker.js.map");
+project.compileTask.spawn(replaceWorker);
```
Chriscbr, line 271: I added the no-op service worker as a JavaScript file directly since the current `service-worker.ts` currently gets treated as special by the `create-react-app` build tools (some plugins try to injecting content into it, and the compiler will yell if I change the file too much). I'd also just rather serve the JavaScript file unminified.

PR #626, opened 2021-11-09
chore: filter ux adjustments

gabewomble:
Implements the following changes:
- Ensure all spaces between filter groups = 16px
- Ensure all spaces between filter label and filter values = 8px
- reduce radio buttons and checkboxes to have size 14x14px
- space between radio buttons and checkboxes = 4px
- Modify question mark icon size from 16x16 to 14x14
- Reduce spacing between tooltip icon and label to 6px

### DIFF
--- a/src/theme/components/Checkbox.ts
+++ b/src/theme/components/Checkbox.ts
@@ -1,0 +1,109 @@
+import { checkboxAnatomy as parts } from "@chakra-ui/anatomy";
+import {
+  mode,
+  PartsStyleFunction,
+  PartsStyleObject,
+  SystemStyleFunction,
+  SystemStyleObject,
+} from "@chakra-ui/theme-tools";
+
+const baseStyleControl: SystemStyleFunction = (props) => {
+  const { colorScheme: c } = props;
+
+  return {
+    w: "100%",
+    transitionProperty: "box-shadow",
+    transitionDuration: "normal",
+    border: "2px solid",
+    borderRadius: "sm",
+    borderColor: "inherit",
+    color: "white",
+
+    _checked: {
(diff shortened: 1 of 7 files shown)


src/theme/components/Checkbox.ts
```diff
@@ -0,0 +1,109 @@
...
+    control: { w: 5, h: 5 },
+    label: { fontSize: "lg" },
+    icon: { fontSize: "0.625rem" },
+  },
+};
+
+const defaultProps = {
+  size: "base",
+  colorScheme: "blue",
+};
```
gabewomble, line 102: With both of these component theme files, the main addition is a `base` size that is applied as the default size. The rest of the file is the existing theme

Chriscbr: This is all front end wizardry to me lol

Chriscbr (commented):
I was trying this locally and it looked to me like the checkboxes are spaced tighter than the radio buttons - just want to double check if this is intended.

Screen Shot 2021-11-09 at 10 14 39 AM

src/theme/components/Checkbox.ts
gabewomble: Haha no worries! Basically what's going on is the `size` prop is similar to a class name that would modify the sizes for a component. Think of it as a js way of doing something like `.btn-sm` / `.btn-md` / `.btn-lg`

gabewomble:
@Chriscbr Good callout - the default spacing on the Radio elements was 8px instead of 4. Also noticed in your screenshot that the radio's `checked` state was misaligned on chrome so I snuck in a fix for that as well

Chriscbr (approved):

PR #627, opened 2021-11-09
chore: search label adjustments

### LINKED ISSUE #587: Search results null state
Song-I's request:
Can we update copy -- instead of “Displaying *0 - 0* of *0* search results for *blah”,* having the copy be something like “There were no results for *blah*. Try a different term.”

### LINKED ISSUE #588: Search Results Page: Fix navigation after clicking the author name or tag
When a user clicks an author name from search results page after scrolling down, take the user back to top of searchresults page, as opposed to preserving location even after the results adjust accordingly.

Same thing above, for when user selects a tag

gabewomble:
Fixes #588 
Fixes #587 

Implements the following changes:
- When there is only a single page of results, display "Showing **x** of **x** results"
- Scroll to top when search results change
- When there are no results, display "There were no results for **X**. Try a different term."

### DIFF
--- a/src/views/SearchRedesign/SearchDetails.tsx
+++ b/src/views/SearchRedesign/SearchDetails.tsx
@@ -16,6 +16,26 @@ const Em: FunctionComponent = ({ children }) => (
   </Text>
 );
 
+const Count: FunctionComponent<{ first: number; count: number; last: number }> =
+  ({ first, count, last }) => {
+    if (!first && last >= count) {
+      return (
+        <>
+          <Em>{count}</Em> of <Em>{count}</Em>
+        </>
+      );
+    }
+
+    return (
+      <>
+        <Em>
+          {count ? first + 1 : count} - {last > count ? count : last}
+        </Em>{" "}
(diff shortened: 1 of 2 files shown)


src/views/SearchRedesign/SearchDetails.tsx
```diff
@@ -25,19 +45,25 @@ export const SearchDetails: FunctionComponent<SearchDetailsProps> = ({
...
-        {count ? first + 1 : count} - {last > count ? count : last}
-      </Em>{" "}
-      of <Em>{count}</Em> {filtered ? "search results" : "constructs"}
+      {hasResults ? (
+        <>
+          Displaying <Count count={count} first={first} last={last} />{" "}
+          {filtered ? "search results" : "constructs"}
+        </>
+      ) : (
+        <>{filtered ? "There were no search results" : "No constructs found"}</>
```
Chriscbr, line 58: I like this detail 👍

Chriscbr (approved):

PR #628, opened 2021-11-09
chore: tab adjustments

gabewomble:
- Sets font weight on selected tab to semibold
- Removes full width underline, sets active underline to have a 4px width

Screen Shot 2021-11-09 at 11 46 34 AM


src/theme/components/index.ts
```diff
@@ -4,19 +4,21 @@ import { Checkbox } from "./Checkbox";
...
 import { Radio } from "./Radio";
+import { Tabs } from "./Tabs";
 import { makeTag } from "./Tag";
 
 interface ComponentsConfig {
   Tag: PackageTagConfig[];
 }
 
-export const makeComponents = (config: ComponentsConfig) => {
+export const makeComponents: any = (config: ComponentsConfig) => {
```
MrArnoldPalmer, line 14: why does this need an `any` now?

gabewomble: Ahh we can probably remove this - I was getting a pretty annoying typescript error in my editor. I gave a basic annotation to suppress the error, but it's definitely not a real solution

Screen Shot 2021-11-10 at 8 50 38 AM

Chriscbr (approved):

PR #640, opened 2021-11-10
feat: keywords filter

### LINKED ISSUE #562: Add keywords filter
To allow users to easily distinguish tags from keywords we should implement a separate "keywords" filter where values can be selected from the list or by clicking on keywords on a card or package profile

gabewomble:
Fixes #562 

Implements a checkbox filter for keywords in the package catalog. Behavior is the same as languages - results that match _none_ of the selected filters will be hidden

Interactions / Behavior:
- Displays first 5 keywords, sorted by occurrence in catalog
- A "More options" / "Fewer options" button is used to show or hide remaining keywords. Only  the top 25 keywords are displayed
- Clicking a keyword on a card will apply the keyword as a filter. If the keyword is not in the top 25 keywords it will be shown at the top of the keyword filter section

TODO:
- [x] Write tests
- [x] Review UX interactions with designer
- [x] Review copy / text
- [x] Support keywords from package page

Screen Shot 2021-11-10 at 1 17 20 PM

Screen Shot 2021-11-10 at 1 17 28 PM

Screen Shot 2021-11-10 at 1 17 34 PM


src/views/Package/PackageHeader/Heading.tsx
```diff
@@ -14,6 +14,10 @@ import { CDKTypeIcon, CDKTypeText } from "../../../components/CDKType";
 import { PackageTag } from "../../../components/PackageTag";
 import { KEYWORD_IGNORE_LIST } from "../../../constants/keywords";
 
+interface TagObject extends PackageTagConfig {
+  isKeyword?: boolean;
+}
```
Chriscbr, line 19: I wonder if we can avoid redefining this interface in multiple places (it's also used in Tags.tsx). The rest of this file also has a lot of similar code that looks as though it can be reused / extracted

Chriscbr (approved):
Looks great! I feel like this will really help with discoverability.

Two comments about the tooltip for the filter:

- I wonder if the first sentence ("Focus the results by choosing one or more keywords reflecting the kind of construct you're looking for") seems a bit redundant? As a user it could feel like I'm being told that grass is green, though I'm cool with leaving this in / seeing if anyone else echoes this.
- It might be helpful to mention in some way that the list is specifically based on the most common keywords and is sorted by frequency. If not in the tooltip, it could be a FAQ question maybe.  Other ideas that could be cool to include as follow up enhancements, curious if any of these sound like they have merit: 
- Is there a way we could include a textbox for users to input a keyword manually?
- How much effort would it be to update the list with the top keywords _given the other filters_? (For example, cdk8s projects might have very different keywords than AWS CDK projects!) This probably requires some further design, but it could be a cool idea to consider. :^)
- Perhaps we could highlight matching keywords on each card in the search result. (Or maybe this would be redundant)

Approving with do-not-merge just to make sure see this comment, nothing blocking though.

src/views/Package/PackageHeader/Heading.tsx
gabewomble: Yeah the whole flow for rendering tags / keywords is an opportunity for code-reuse that I can add in a follow-up!

gabewomble:
@Chriscbr All good ideas - a few of which were taken into consideration!

- For the tooltip text, I got this from our technical writer so I tend to trust the experts even though I agree feels pretty obvious
- As for explaining the ordering - it's a good callout. If we were able to display counts next to them it would solve the issue, though it has it's own UX / implementation challenges (which is related to one of your enhancement suggestions). I chatted with Song-I about this when reviewing the work and she suggested we stick with this for the time being

For the enhancements:
- Textbox search - definitely possible, just a feature of it's own to create the component
- Top keywords based on other filters is the tricky one, because the behavior should be consistent with the other filters. If we had a filter panel that had a top-down flow where users added filters rather than them always being present, it would definitely be doable. This is something we could tackle post GA as it would likely be somewhat significant work!
- Highlighting matched keywords would also be nice, but I do think it could be redundant. If we went with that approach I would definitely like to give the same treatment to search query results, though that would be a large effort as well

PR #643, opened 2021-11-11
fix: multiple of same package appears in search results

### LINKED ISSUE #585: Add major version to the package card and package page
Some packages have several major versions. Their cards then look as clone of each other unless we add the major version. 

In addition, it's a helpful metadata for the package page.

Chriscbr:
Fixes #585

Instead of showing the major version of a package (or major version of the CDK framework) to the package cards, we filter out packages appearing multiple times with different major versions from search results, always preferring to show the highest major version of the package that satisfies the search filters.

For example, if you search for `@aws-cdk/cx-api`, by default only one card is shown, that links to `@aws-cdk/cx-api@2` that supports CDK v2. But if you apply a filter for AWS CDK v1, then `@aws-cdk/cx-api@1` is shown.


src/api/catalog-search/catalog-search.ts
```diff
@@ -223,6 +225,25 @@ export class CatalogSearchAPI {
...
+   * one with the latest version.
+   */
+  private dedup(results: CatalogSearchResults): CatalogSearchResults {
+    const dedupedResults: Map<string, ExtendedCatalogPackage> = new Map();
+
+    results.forEach((pkg) => {
+      const maybePkg = dedupedResults.get(pkg.name);
+
+      // include the result if the package is new, or it has a higher major version
+      if (!maybePkg || (maybePkg && maybePkg.major < pkg.major)) {
```
eladb, line 239: This assumes that the latest version is the "most major" version but that's not necessarily true. We can't assume that.

Here's a concrete example that @addihorowitz and I ran into:

The latest version of [cdk-dynamo-table-viewer](https://www.npmjs.com/package/cdk-dynamo-table-viewer) is v0.1.142, while the largest major version is v3.1.2. You'll also notice that 3.1.2 does not have a description (which was added later).

The download numbers you get from npm are only for the last version published under the `latest` dist tag, so if we are deduplicating, we should only show this version and not just the latest major.

eladb (changes requested):
See comment about latest versus last-major.

src/api/catalog-search/catalog-search.ts
Chriscbr: I see, yeah this is a great callout. Looks like what we should probably do is track the NPM dist tags in our backend, or at the minimum try to label in the catalog which major version is tied to "latest" so that package can take priority during this de-duplication.

This is probably a bit of logic that we could tack on during catalog building or ingestion, though the truth is that it's technically possible for library authors to change which version is tagged with "latest" at any time. So the most reliable way to ensure this data doesn't get out of sync would be to update it whenever we ingest a new package AND on some periodic schedule. (which I think implies storing the data someplace besides the catalog). But this edge case is probably small enough that we could defer implementing this "periodic refresh" logic until a later date. 

It would be nice if we could "cheat" by showing the most recently published version as a proxy for the "latest"-tagged version, but that will produce bad results for packages where multiple tags (like "latest" and "next") are both being actively published to.

For now I'll mark this PR as a draft because of the concerns you mentioned.

eladb: Curious, when you get download numbers from npm, do you have a version indication of the module or is it just name?

Chriscbr: It's just by package name. There's no publicly available API for getting download numbers per version.

Chriscbr: > The download numbers you get from npm are only for the last version published under the latest dist tag, so if we are deduplicating, we should only show this version and not just the latest major.

Sorry, I missed this part of your first comment. To clarify ,the npm download numbers we're using are for all versions, not for the "latest" version. For example, https://www.npmjs.com/package/@aws-cdk/cx-api has > 1M downloads (which is also reflected on Construct Hub) but only when summed over all versions.

Following on my earlier comment, I'm starting to think having logic dependent on the "latest" tag could be tricky because it's possible the "latest" tag could point to a package version that hasn't been ingested OR a version that can't be ingested OR a version that has been deny listed. And there's no way to identify what was previously tagged as "latest" in the past. It's possible this is still the correct thing to do, but there's more edge cases than expected...

gabewomble: I believe our best bet would be to just show the most recently updated version until we can support the _latest_ tag. This will at least be pretty consistent with how npm behaves. IMO the above scenario feels like a bit of an edge case, and at the end of the day, we need to show a single result that links to a page that has a version dropdown rather than potentially confusing users with multiple near-identical results. @eladb @Chriscbr Let me know what you think of this.

Chriscbr: I'm OK with this. This will be a little bit weird for packages that publish to multiple tags, but I think it's still a net better experience. Updated the PR.

src/api/package/packages.ts
```diff
@@ -11,6 +11,7 @@ export interface Author {
 export interface CatalogPackage {
   name: string;
   languages?: Partial<Record<Language, Record<string, unknown>>>;
+  major: number;
```
eladb, line 14: Is this needed?

src/api/catalog-search/catalog-search.ts
```diff
@@ -191,6 +191,8 @@ export class CatalogSearchAPI {
       results = this.sort(results, sort);
     }
 
+    results = this.dedup(results);
```
eladb, line 194: No test?

src/api/catalog-search/catalog-search.ts
```diff
@@ -283,6 +285,27 @@ export class CatalogSearchAPI {
...
   }
 
+  /**
+   * De-duplicates packages that appear multiple times, keeping the
+   * most recently published one.
+   */
+  private dedup(results: CatalogSearchResults): CatalogSearchResults {
+    const dedupedResults: Map<string, ExtendedCatalogPackage> = new Map();
+
+    results.forEach((pkg) => {
```
eladb, line 295: Style nit: prefer `for-if` instead of `forEach()`. It's easier to read and maintain in the long term. For example, it is not possible to `break` a `forEach()` loop.

src/api/package/packages.ts
eladb, line 14: Is this needed?

eladb (changes requested):

src/api/package/packages.ts
gabewomble: This is actually a fix on the type - `CatalogPackage` represents data we get from the backend, so this annotation was previously missing

Chriscbr:
Several tests needed tweaking since they tend to assume that no de-duplication goes on. Initially I thought that was OK, but now I'm realizing that for #679 I will actually need to have a way to retrieve un-de-duplicated results, so I'm going to switch "dedup" into being an optional search flag.

gabewomble:
@Chriscbr I think it makes sense to `dedup` on search by default with a flag to disable it, but IMO it would be nice to have a method on the search such as `findByName` which could just return a list or map of packages that exact match the name

Chriscbr:
@gabewomble Good suggestion, I think I'll add a method like that in my other PR.

Chriscbr:
@eladb @gabewomble Is there anything blocking this UX fix?

gabewomble:
Nothing on my end 👍

src/api/catalog-search/catalog-search.ts
```diff
@@ -92,6 +92,7 @@ export interface CatalogSearchParams {
   query?: string;
   filters?: CatalogSearchFilters;
   sort?: CatalogSearchSort;
+  dedup?: boolean;
```
gabewomble, line 95: I think I'd prefer this not to be optional so that behavior is consistent across the board. As is, we'll still get duplicated results from search suggestions. LMK what you think, we could also default this to true as an alternative

gabewomble (approved):

PR #650, opened 2021-11-11
feat: search relevance tuning

### LINKED ISSUE #644: Must have search algorithm improvements
Many relevant results are not retrieved

- [ ] #645 
- [ ]  Search algorithm: Exact match should get higher priority #614

gabewomble:
Fixes #644 

This work will need a bit of testing to validate as it relies on a fair amount of trial and error. Will add a better write up as I get closer to a point of completion

Note: This work is not yet ready as there are some edge cases where certain names don't match the way they should


MrArnoldPalmer (approved):
added `do-not-merge` in case you think this still needs tweaking. However I believe this to be clearly superior to the current search so I'm ready to move forward and tackle additional issues we find separately.

gabewomble:
Yep! We can follow up with additional improvements as needed!

PR #652, opened 2021-11-12
fix: md code blocks breaking navigation

MrArnoldPalmer:
If a line within a fenced code block in a markdown document begins with `#`, then our parsing would mistakenly identify it as a header. Prevent that from occurring by parsing out fenced code blocks before parsing out header lines.

NOTE: currently does not support [indented or "unfenced" code blocks](https://github.github.com/gfm/#indented-code-blocks).


gabewomble:
Marking as draft to prevent merges - found an edge case on `cdk-iam-floyd` that crashes from a recursion loop

gabewomble (approved):

gabewomble:
Had to give this a little extra testing time to be sure, but it works! Nice job 👍

PR #656, opened 2021-11-12
feat: show all keywords on package page

gabewomble:
Fixes https://github.com/cdklabs/construct-hub/issues/538

We can now display more keywords on the package header! This is using flex gap which is supported on Firefox, Chrome, and Safari meaning we don't have to hack a solution when items wrap

Screen Shot 2021-11-12 at 2 18 18 PM
Screen Shot 2021-11-12 at 2 18 25 PM

### DIFF
--- a/src/views/Package/PackageHeader/HeaderContainer.tsx
+++ b/src/views/Package/PackageHeader/HeaderContainer.tsx
@@ -24,12 +24,12 @@ const lgGridAreas = makeGridAreas(
 export const HeaderContainer: FunctionComponent = ({ children }) => (
   <Grid
     columnGap={{ md: 6, lg: 10 }}
-    pt={{ base: 3, md: 6 }}
+    pt={{ base: 3, lg: 6 }}
     px={{ base: 5, md: 6, lg: 10 }}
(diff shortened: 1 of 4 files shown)


src/views/Package/PackageHeader/HeaderContainer.tsx
```diff
@@ -24,12 +24,12 @@ const lgGridAreas = makeGridAreas(
...
   <Grid
     columnGap={{ md: 6, lg: 10 }}
-    pt={{ base: 3, md: 6 }}
+    pt={{ base: 3, lg: 6 }}
     px={{ base: 5, md: 6, lg: 10 }}
-    rowGap={{ base: 4, md: 0 }}
-    templateAreas={{ base: baseGridAreas, md: lgGridAreas }}
-    templateColumns={{ base: "1fr", md: "1fr auto 15rem" }}
-    templateRows={{ md: "auto 1fr" }}
+    rowGap={{ base: 4, lg: 0 }}
```
gabewomble, line 29: This updates the heading layout to break to the mobile version earlier

Chriscbr (approved):

PR #674, opened 2021-11-15
feat: add tags to search indexing

### LINKED ISSUE #666: The search algorithm should also run the search on the packages' tags

gabewomble:
Fixes #666 

Adds tags to search indexing. Also updates the catalog estimate in the search placeholder to round down, and resets sorting to relevance when a user types a new input

### DIFF
--- a/src/api/catalog-search/catalog-search.ts
+++ b/src/api/catalog-search/catalog-search.ts
@@ -36,6 +36,10 @@ const INDEX_FIELDS = {
     name: "scope",
     boost: 5,
   },
+  TAG_NAMES: {
+    name: "tagNames",
+    boost: 2,
+  },
 } as const;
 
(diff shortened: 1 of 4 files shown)


src/api/catalog-search/catalog-search.ts
```diff
@@ -109,18 +114,42 @@ export class CatalogSearchAPI {
...
+          authorName = author;
+        } else {
+          if (author?.name) {
+            authorName = author.name;
+          }
 
+          if (author?.email) {
+            authorEmail = author.email;
+          }
+        }
```
gabewomble, line 140: I've moved all this data mapping to when we define the catalog map, rather than when we pass it to the index to prevent any confusion / mutability related bugs

Chriscbr (approved):

PR #679, opened 2021-11-16
feat: select major version on package page

### LINKED ISSUE #585: Add major version to the package card and package page

Chriscbr:
Closes #585

~Screenshots:~ See updated screenshots in comments
Screen Shot 2021-11-15 at 4 08 01 PM
Screen Shot 2021-11-15 at 4 08 18 PM


Chriscbr:
Updated screenshots

Screen Shot 2021-11-16 at 2 27 41 PM
Screen Shot 2021-11-16 at 2 28 17 PM
Screen Shot 2021-11-16 at 2 29 28 PM

src/api/catalog-search/catalog-search.ts
```diff
@@ -182,10 +183,11 @@ export class CatalogSearchAPI {
     query?: string;
     filters?: CatalogSearchFilters;
     sort?: CatalogSearchSort;
+    exactMatch?: boolean;
```
gabewomble, line 186: I don't think this is the way to go. IMO it would make more sense to implement this as a standalone method rather than expanding the surface area of the `search` api and related hooks. I.E `CatalogSearchAPI.findByName`

This would be exposed to our UI code
via `useSearchContext` like so:

```tsx
const searchAPI = useSearchContext();

const versions = searchAPI.findByName("@aws-cdk/core")
```

src/api/catalog-search/catalog-search.ts
```diff
@@ -226,6 +228,10 @@ export class CatalogSearchAPI {
     return refs.reduce((packages, { ref }) => {
       const pkg = this.map.get(ref);
 
+      if (exact === true && pkg?.name !== query) {
+        return packages;
+      }
```
gabewomble, line 233: See above comment

src/views/Package/PackageHeader/SelectVersion.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const { push } = useHistory();
+
+  const packageMajorVersions = packages.results.map((pkg) => ({
+    major: pkg.major,
+    version: pkg.version,
+  }));
+  packageMajorVersions.sort((a, b) => a.major - b.major);
+  const defaultMajor = packageMajorVersions.find(
+    (mv) => mv.version === version
+  );
```
gabewomble, line 25: ⚠️ I would compute this in a single `useMemo` effect. The current implementation means that `packageMajorVersions` will be recomputed on every render

gabewomble (changes requested):
Mostly looks good, though I had a few thoughts on the changes to the CatalogSearch api

src/api/catalog-search/catalog-search.ts
Chriscbr: Agreed, thanks for the suggestion.

src/views/Package/PackageHeader/SelectVersion.tsx
Chriscbr: Updated.

Chriscbr:
@gabewomble Thanks for the patience as I keep iterating on the design :).

Screenshot of newest version of the UI : 

Screen Shot 2021-11-16 at 4 04 49 PM

This should also be pretty easy to update once we have a full list of versions available from the backend.

src/components/CDKType/CDKType.tsx
```diff
@@ -9,10 +9,11 @@ import { CDKType, CDKTYPE_RENDER_MAP } from "../../constants/constructs";
 
 interface CDKTypeIconProps extends ImageProps {
   name?: CDKType;
+  majorVersion?: number;
 }
 
 export const CDKTypeIcon = forwardRef<CDKTypeIconProps, "img">(
-  ({ name, ...props }, ref) => {
+  ({ name, majorVersion, ...props }, ref) => {
```
gabewomble, line 16: Curious about this addition here

src/api/catalog-search/catalog-search.ts
```diff
@@ -234,6 +234,20 @@ export class CatalogSearchAPI {
     }, new Map() as CatalogSearchResults);
   }
 
+  /**
+   * Performs a Search against the catalog and returns an array of all packages matching the query.
+   */
+  public findByName(query: string): ExtendedCatalogPackage[] {
+    const results = this.query(query);
```
gabewomble, line 241: I think we can skip the `query` function and do something like

```tsx
const results = [...this.map.values()].filter(pkg => pkg.name === query));
```

src/components/CDKType/CDKType.tsx
Chriscbr: This is an unrelated bug fix - I saw a console error that said (paraphrasing) "majorVersion can't be passed as a prop to this IMG element" and I figured out it's because it was getting grouped into `...props` here.

gabewomble (approved):
Approving with do-not-merge in case @Chriscbr needs time to add anything else. Awesome work 🚀

Chriscbr:
@mergifyio refresh

PR #682, opened 2021-11-16
feat: categories section in home page

eladb:
Add a "categories" section in the home page with a set of buttons that lead to keyword searches.

If the backend config includes a `categories` section, it will be used to define the set of categories and keywords. If it is absent, a default set will be included.

The new design also switches the colors of the Featured and Tabs section in the home page.

Desktop:

Screen Shot 2021-11-16 at 12 18 55

Mobile:

Screen Shot 2021-11-16 at 12 19 49


eladb:
@addihorowitz what do you think about the texts? In the meantime I've used the ones from the design but I am not sure about them.

src/views/Home/CDKTypeTabs.tsx
```diff
@@ -30,6 +30,7 @@ interface PackageTabProps {
 const PackageTab: FunctionComponent<PackageTabProps> = ({ cdkType, data }) => {
   return (
     <Tab
+      _selected={{ borderBottom: "4px solid #AED5FF" }}
```
gabewomble, line 33:
```suggestion
      _selected={{ borderBottomColor: "white" }}
```

src/views/Home/CDKTypeTabs.tsx
```diff
@@ -91,8 +92,6 @@ export const CDKTypeTabs: FunctionComponent = () => {
 
   return (
     <Flex
-      bg="white"
-      color="blue.800"
```
gabewomble, line 95: I would continue setting the `color` here and allowing inheritance to set it for the text below

src/views/Home/CDKTypeTabs.tsx
eladb: The design is to use a different color. Should I use white?

gabewomble (commented):
I think the UI for this is really nice, but I have a few UX concerns that leave me hesitant to merge this as is

- Not all keywords have search results, which would be a frustrating experience for a user
- I believe this feature should leverage the tag `api` which would allow us to define these use cases as tags (an initial implementation could tag based on keywords). This would also give us a filter section for these "Use Cases" for free (it's what we do for Publisher tags). By also having the filter section on search, users would not be required to go back & forth between the home and search pages to check out different categories

src/views/Home/CDKTypeTabs.tsx
gabewomble: My bad - I wasn't aware there were design changes outside of this new section - disregard

eladb:
> I believe this feature should leverage the tag api which would allow us to define these use cases as tags (an initial implementation could tag based on keywords). This would also give us a filter section for these "Use Cases" for free (it's what we do for Publisher tags). By also having the filter section on search, users would not be required to go back & forth between the home and search pages to check out different categories

@MrArnoldPalmer I've discussed this with @gabewomble and to my understanding if we want to use tags for this it means that every time we add a category we will need to reprocess the entire catalog. If that's the case, then I rather keep this simple and based just on keywords. Any strong opinions here?

src/views/Home/CDKTypeTabs.tsx
eladb: Cool, so leaving as is to align with design.

MrArnoldPalmer:
@eladb yeah I think we can add support for categories to reference tags later and just make them do keyword searches when they aren't pointing to a computed tag. These categories feel broad enough that packages should be found where relevant. Maybe we should make them just "?q=categoryname" so the package author doesn't have to have a keyword actually matching the category and we can just do a general search?

eladb:
Changing to draft for now until we finalize copy & clean up categories.

gabewomble:
@eladb @MrArnoldPalmer's suggestion is actually a really great way to solve this without tags. Our search queries look at keywords as well as other information on packages so we will be able to pull results for relevant libraries that may not have the selected keyword.

src/views/Home/Categories.tsx
```diff
@@ -0,0 +1,86 @@
...
+              colorScheme="gray"
+              size="md"
+              style={{ boxShadow: "0px 4px 4px rgba(73, 73, 73, 0.63)" }}
+            >
+              <Link
+                href={getSearchPath({ keywords: [category.searchKeyword] })}
+              >
+                {category.title}
+              </Link>
+            </Button>
```
gabewomble, line 80: This actually needs to be a `NavLink` instead of a regular href because it links to another page in-app

`import { NavLink } from "../../components/NavLink";`

```tsx
<Button
              as={NavLink}
              color="blue.800"
              colorScheme="gray"
              size="md"
              style={{ boxShadow: "0px 4px 4px rgba(73, 73, 73, 0.63)" }}
              to={getSearchPath({ query: category.searchKeyword })}
            >
              {category.title}
            </Button>
```

The reason we want to use the NavLink component is that it won't trigger a browser refresh which is unnecessary as we serve a single-page application.

We can use the `as` prop in the button to render the `NavLink` like a button - a really nice feature from chakra-ui

src/views/Home/CDKTypeTabs.tsx
eladb: TIL!

src/views/Home/Categories.tsx
eladb: TIL

eladb:
> @eladb @MrArnoldPalmer's suggestion is actually a really great way to solve this without tags. Our search queries look at keywords as well as other information on packages so we will be able to pull results for relevant libraries that may not have the selected keyword.

OK, switching to `q=?`

eladb:
@gabewomble After changing to `?q=X`, all categories besides "Cost Management" now have packages in them, so I removed this category for now.

@addihorowitz waiting for copy editing and we can merge this!

gabewomble (approved):
Really dig this feature, nice work!

addihorowitz (commented):
Approved. Let's improve the text later this week and integrate the solutions with the new tags we're going to add

eladb:
@gabewomble @MrArnoldPalmer I've discussed the change to `?q=` with @addihorowitz and we are not sure that this is the correct behavior (at least not for all categories). So what I did is that a category now simply defines a `url` instead of `searchKeywords`. This will allow us to configure in the backend exactly what URL will be behind each category. For example, we could decide that some categories are bound to keywords, some to general query, etc.

PR #683, opened 2021-11-16
chore: search result improvements

RomainMuller:
- Sort by date only uses the date portion of the timestamp, and uses
  alphanumeric ordering as a tie-breaker. The "Recently Updated" section
  of the landing page continues to do "full-timestamp-only" ordering.

  * Fixes https://github.com/cdklabs/construct-hub-webapp/issues/662

- The `<Time>` component now renders the ISO-8601 UTC timestamp as the
  `title` property of the HTML, which makes browsers render a tooltip
  when the user leaves the mouse over the item for long enough.

- The package version is now rendered on the search result cards
  (including on the landing page), so that users can easily determine
  what version each card is about (this is in particular useful for
  reasoning over multiple cards with the same package name, such as
  those for `@aws-cdk/cx-api`).

  * Fixes https://github.com/cdklabs/construct-hub-webapp/issues/585


src/api/catalog-search/catalog-search.ts
```diff
@@ -155,7 +155,22 @@ export class CatalogSearchAPI {
...
+        const rDate = new Date(right.metadata.date);
+
+        if (lDate === rDate) {
+          return 0;
+        } else if (lDate < rDate) {
+          return 1;
+        } else {
+          return -1;
+        }
+      }
```
gabewomble, line 172: Why would we not use the same sort function here as we do for results? It would allow us to stick with one tested solution

src/api/catalog-search/catalog-search.ts
```diff
@@ -276,13 +291,13 @@ export class CatalogSearchAPI {
    */
   private sort(
     results: CatalogSearchResults,
-    strategy: CatalogSearchSort
+    strategy: CatalogSearchSort | SortFunction
   ): CatalogSearchResults {
     if (strategy) {
+      const sortFunction =
+        typeof strategy === "function" ? strategy : SORT_FUNCTIONS[strategy];
```
gabewomble, line 298: See above comment

src/api/catalog-search/catalog-search.ts
RomainMuller: When we show a short-list of "recently updated" packages, I'd wanna look at them strictly by publish date... Otherwise we'll show the same 4 packages for the whole day, and this felt wrong.

gabewomble: @RomainMuller I think this is a good argument why we should not compare off date only. Additionally this will lead to inconsistent results for users in different timezones which I would like to avoid

src/api/catalog-search/util.ts
```diff
@@ -11,13 +12,31 @@ type FilterFunctionBuilder<T> = (
...
-    const d1 = new Date(p1.metadata.date);
-    const d2 = new Date(p2.metadata.date);
+    // Sorting only by the DATE portion of the timestamp. ISOString is YYYY-MM-DD'T'HH:mm:ss.SSSZ
+    const d1 = new Date(p1.metadata.date).toISOString().split("T")[0];
+    const d2 = new Date(p2.metadata.date).toISOString().split("T")[0];
+
+    // If they're equal, fall back to alphanumerical ordering.
     if (d1 === d2) {
-      return 0;
+      return getStrSort(!isAscending)(p1, p2);
```
gabewomble, line 39: Thought I left a comment here, but I strongly believe we should avoid sorting just by date - this breaks the newest / oldest flow quite a bit because as your above comment said, all results will just be alphabetical for the same day and it's going to be very odd from a user perspective

gabewomble (changes requested):
I don't believe the changes to date sorting will lead to a good user experience. I dig the other changes though (version sort, version on card)

PR #687, opened 2021-11-16
feat: adds highlights to package page

### LINKED ISSUE #639: Community/AWS/HashiCorp/DataDog badge is missing from the package page
Should be on the right side with the metadata of the package

### LINKED ISSUE #646: Package header is broken
- [x] #633 
- [x] #639
- [x] #580 
- [x] #585 
- [x] #669

gabewomble:
Fixes #639 

The core feature is adding Highlights to the package page, but this PR also has a few additional changes:
- Created utils for mapping package tags, keywords, and highlights. We were previously re-using code in multiple places, which isn't ideal for obvious reasons.
- Improved the Details section to only show two items when collapsed, but now it is open by default

Screen Shot 2021-11-16 at 12 35 50 PM


src/util/package.ts
```diff
@@ -0,0 +1,70 @@
...
+      keyword: {
+        label,
+      },
+    }));
+};
+
+/**
+ * Maps packageTags and keywords to a list of TagObjects, using mapPackageTags and mapPackageKeywords
+ */
+export const mapPackageTagsAndKeywords = ({
```
Chriscbr, line 62: Same comment here, could make sense just to be explicit here e.g. `intoTagObjects` or `convertIntoXxx`... not sure.

Chriscbr (approved):
Had some suggestions about naming, but nothing blocking. LGTM

PR #693, opened 2021-11-17
fix: nav items broken for submodules

### LINKED ISSUE #692: submodule navigation bar is broken
When I load a module that includes submodules (such as `aws-cdk-lib`) and go to a specific submodule and click on a navigation item, it jumps back to the root module.

eladb:
Add a `submodule=` query parameter if navigation item is within a submodule.

Fixes #692


RomainMuller:
Seems to be missing tests?

eladb:
@RomainMuller added test

RomainMuller (approved):

Issue #667, opened 2021-11-15
Package card/page: Tags and keywords should be unique


addihorowitz:
We should display keywords and/or tags only once per package and avoid displaying duplicate indications. 

For example if a package has the keyword 'AWS S3' and there's a tag named 'AWS S3' only one of them should be displayed.

eladb:
Resolved by #694

PR #694, opened 2021-11-17
feat: search in tag-based keywords

### LINKED ISSUE #684: Add tags to the Search Results page
Add the tags to the keywords filter on the search results page. 

Once a user pressed a category on the home page, the user will be redirected to the search results page and this tag will be clicked. The relevant search results that have this tag will be displayed. 

Please verify with @gabewomble if the search algorithm already supports searching over tags.

eladb:
When querying `keywords=`, then include keywords that come from tags and not just keywords published by the package.

Also, include keywords that originate from package tags when rendering the filter panel.

Resolves #684


src/__fixtures__/catalog.json
```diff
@@ -1560,7 +1560,20 @@
         },
         "date": "Mon, 20 Sep 2021 19:04:49 GMT",
         "packageLinks": {},
-        "packageTags": []
+        "packageTags": [
```
eladb, line 1563: Added some synthetic tag-based keywords here for testing purposes.

gabewomble (approved):
Looks solid to me 👍 
@addihorowitz @eladb I believe this also closes https://github.com/cdklabs/construct-hub-webapp/issues/667

Issue #670, opened 2021-11-15
Clicking on items in the navigation area doesn't work


addihorowitz:
https://constructs.dev/packages/cdk-construct-simple-nat/v/0.1.125?lang=typescript#whats-the-difference-between-ec2-nat-instancesnatinstances-and-nat-instances-created-by-this-construct

Try to click the items under the FAQ

RomainMuller:
This appears to be due to link-in-link here, making the heading anchor not be what it's supposed to be.

PR #695, opened 2021-11-17
fix: incorrect links in nav tree

RomainMuller:
When a markdown heading includes in-line links (`[label](href)` or
`[label][ref]`), the label of the item in the nav tree was incorectly
rendering raw markdown, and the anchor ID was not correct.

This addresses both of these issues:

- The anchor ID is now computed based on the recursive text elements
  from the DOM node (previously, it would not include any non-text
  child).

- The nav tree title now removes the markdown link decorations, and
  leaves only the label in the title & id.

- The nav tree title now removes custom anchors (`<a name="...">`)
  and leaves only the relevant text to compute the title & id.

Fixes https://github.com/cdklabs/construct-hub-webapp/issues/670


RomainMuller:
BTW - not too sure how to test this... Pointers/ideas welcome!

gabewomble:
> BTW - not too sure how to test this... Pointers/ideas welcome!

I believe there is a suite of tests regarding markdown parsing @MrArnoldPalmer wrote, I think that would be a good place to add tests for this!

gabewomble:
It looks like nav items are still broken for `iam-floyd`, not sure if that case was intended to be solved with this PR, but for the case you linked, the changes work as expected!

RomainMuller:
> It looks like nav items are still broken for `iam-floyd`

This one is also fixed now... It was a different cause.

src/views/Package/util.ts
```diff
@@ -83,8 +83,17 @@ const getHeaderAttributes = (hdr: string): { id: string; title: string } => {
...
-  const [_, rawTitle] = /^#*\s*([^<]+?)\s*(?:<|$)/.exec(hdr) ?? [];
-  const wEmoji = rawTitle.replace(
+  const [_, rawTitle] =
+    /^#*\s*(?:<a\s.*?(?:\/>|<\/a>))?\s*([^<]+?)\s*(?:<|$)/.exec(hdr) ?? [
+      hdr,
+      hdr,
+    ];
+  const noLinkTitle = rawTitle.replace(
+    // Matches [label](text) and [label][text], replaces with just label.
+    /\[([^\]]+)\](?:\([^)]+\)|\[[^\]]+\])/g,
```
gabewomble, line 93: I'm convinced you dream in RegExp 😛

gabewomble (approved):
Nicely done 😎

src/views/Package/util.ts
RomainMuller: I dream electric sheep.

PR #696, opened 2021-11-17
chore: update keyword ignore list

### LINKED ISSUE #661: Keywords fixes
1. All (most) keywords should be visible on the package card on the search results page (not only the package page)
2. Trivial keywords (like cdk, aws-cdk, cfktf, cdk8s) should be removed from the keywords filter
3. cdktf should not be displayed as a keyword (just like aws-cdk, cdk etc.)

Chriscbr:
Fixes #661

Updates the keyword ignore list, and increases the number of tags allowed on package cards.


src/components/PackageCard/Tags.tsx
```diff
@@ -32,7 +32,7 @@ export const Tags: FunctionComponent = () => {
   return (
     <>
       {tags
-        .slice(0, 3)
+        .slice(0, 10)
```
Chriscbr, line 35: Our package cards currently don't "wrap" the tags area, so if there are 10 tags but only 3 can fit on the first line because of a narrow screen, then only 3 will appear. So this change should be relatively safe / won't cause the cards to become too vertically bloated.

gabewomble (approved):

PR #697, opened 2021-11-17
feat: add Contribute page

### LINKED ISSUE #676: Implement Contribute page
* Add the Contribute link we originally planed on the header.
*  Create a new page path should be: /contribute
* The page should look like the FAQ page but with the title: 'How to contribute?'
* The questions and answers for this page can be found in the Content Review doc.

gabewomble:
Fixes #676 

Screenshot 2021-11-17 at 13-57-55 Contribute - Construct Hub


addihorowitz (commented):
Awesome!

src/components/Header/MobileNav.tsx
```diff
@@ -50,6 +52,10 @@ export const MobileNav: FunctionComponent<MobileNavProps> = ({
                   },
                 ]}
               />
+
+              <NavLink color="blue.800" fontWeight="500" to={ROUTES.CONTRIBUTE}>
+                Contribute
+              </NavLink>
```
Chriscbr, line 58: The mobile nav link does not match the font color / size of the other top-level headers ("Getting Started" and "Documentation")

src/views/Contribute/Contribute.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+      <Accordion allowMultiple allowToggle defaultIndex={[0]}>
+        <FAQItem question="How can I add my construct to Construct Hub?">
+          <Stack spacing={4}>
+            <Text>
+              Constructs intended for Construct Hub must be published to the{" "}
+              <FAQLink href="https://www.npmjs.com/">NPM registry</FAQLink>{" "}
+              under a permissive license (such as Apache, BSD, EPL, MPL-2.0, ISC
+              and CDDL or MIT) and annotated with a keyword recognized by
+              Construct Hub (awscdk, cdk8s, or cdktf).
```
Chriscbr, line 32: It looks like this question is the same as one in the FAQ, but the answer text doesn't match (the list of licenses is different, and it's missing a link on "file an issue against our GitHub repository"). The text should be fixed to match (or we could remove the question from the FAQ and only keep the copy here).

src/views/Contribute/Contribute.tsx
```diff
@@ -0,0 +1,311 @@
...
+        </FAQItem>
+
+        <FAQItem question="What are the best practices for improving discoverability of a construct library on Construct Hub?">
+          <Stack spacing={4}>
+            <Text>
+              To improve your construct library’s visibility on Construct Hub,
+              follow these best practices:
+            </Text>
+            <OrderedList pl={8} sx={{ li: { mt: 4 } }}>
+              <ListItem>
```
Chriscbr, line 82: Can we decrease the margin between the paragraph and the first list item? see screenshot:

Screen Shot 2021-11-17 at 2 35 17 PM

src/components/FAQ/FAQSections.tsx
```diff
@@ -0,0 +1,7 @@
+import { Accordion, AccordionProps } from "@chakra-ui/react";
+import type { FunctionComponent } from "react";
+
+export const FAQSections: FunctionComponent<AccordionProps> = ({
+  children,
+  ...accordionProps
+}) => <Accordion {...accordionProps}>{children}</Accordion>;
```
Chriscbr, line 7: I wonder if it would make sense to make "allowMultiple" enabled by default in this component?

src/views/Contribute/Contribute.tsx
```diff
@@ -0,0 +1,311 @@
...
+    meta={{
+      description: "Learn about contributing to Construct Hub.",
+      title: "Contribute",
+    }}
+    pageName="contribute"
+  >
+    <FAQ>
+      <FAQHeading>Contributing to Construct Hub</FAQHeading>
+
+      <Accordion allowMultiple allowToggle defaultIndex={[0]}>
```
Chriscbr, line 24: Would it make sense to use `FAQSections` here? Also I think `allowToggle` is enabled by default (could be wrong)

Chriscbr (changes requested):

src/views/Contribute/Contribute.tsx
gabewomble: Oh yeah, this shouldn't have so much margin on the top lol

src/views/Contribute/Contribute.tsx
gabewomble: Not really, since `FAQSections` is the parent component we use on `views/FAQ` to give a higher-level heading such as `General`

src/components/Header/MobileNav.tsx
gabewomble: Screen Shot 2021-11-17 at 3 48 07 PM

src/components/Header/MobileNav.tsx
```diff
@@ -50,6 +52,15 @@ export const MobileNav: FunctionComponent<MobileNavProps> = ({
...
               />
+
+              <NavLink
+                color="blue.800"
+                fontSize="1rem"
+                fontWeight="bold"
+                to={ROUTES.CONTRIBUTE}
+              >
+                Contribute
+              </NavLink>
```
Chriscbr, line 63: super small nitpick, but can we make the tap area for this link span the full horizontal width for mobile users?

Chriscbr (changes requested):

src/components/Header/MobileNav.tsx
gabewomble: Yeah this link will look / feel more like the buttons it is placed next to

Screen Shot 2021-11-18 at 8 53 38 AM

Chriscbr (approved):

PR #698, opened 2021-11-17
feat: updates card layout

gabewomble:
This is _hopefully_ the last card-specific update 🙂 

After some discussion with our designer, we're making a few small tweaks to the layout to allow this component to display more comfortably across a wide variety of viewports and datasets.
- Metadata on the right now has a border to clearly separate it from title / description
- Tag limit is increased
- Editor's note will display in-place of descriptions on the featured section
- Metadata items labels have been removed to make better use of space

Screen Shot 2021-11-17 at 3 10 09 PM
Screen Shot 2021-11-17 at 3 10 13 PM
Screen Shot 2021-11-17 at 3 10 21 PM
Screen Shot 2021-11-17 at 3 10 28 PM
Screen Shot 2021-11-17 at 3 11 36 PM
Screen Shot 2021-11-17 at 3 11 41 PM


src/components/PackageCard/Heading.tsx
```diff
@@ -10,14 +10,15 @@ import { Link } from "react-router-dom";
 import { useLanguage } from "../../hooks/useLanguage";
 import { getPackagePath } from "../../util/url";
 import { CDKTypeIcon, CDKTypeText } from "../CDKType";
+import { EditorsNote } from "./EditorsNote";
 import { usePackageCard } from "./PackageCard";
 import testIds from "./testIds";
 
 export const Heading: FunctionComponent = () => {
```
Chriscbr, line 17: This isn't part of your changes, but I did want to raise the question of whether we should change the card title to use `wordBreak="normal"` instead of `wordBreak="break-all"`, so that it prefers to break on dashes instead of in the middle of words. I feel like this could improve the readability in some areas, but I wonder if you or anyone else has found any edge cases where its problematic.

Chriscbr (commented):
Looks great, just had one minor question.

src/components/PackageCard/Heading.tsx
gabewomble: Good callout, updated 👍

Chriscbr (approved):

PR #702, opened 2021-11-18
fix: duplicate keywords/tags appear on packages

### LINKED ISSUE #701: De-duplicate keywords
Keywords on package cards (some of which are _actual keywords_ generated by authors, and some of which are tags that we have added) may be duplicated because of overlap between keywords and tags. ~"Database" and "database" should be de-duplicated into just "Database".~

Related to #645

Chriscbr:
Fixes #701

There are three sources of duplicate keywords in both package cards and on the package page:

- The first source of duplicate keywords is that https://github.com/cdklabs/construct-hub-webapp/pull/694 changes the search index to treat tags as keywords on the search page (specifically on this line): https://github.com/cdklabs/construct-hub-webapp/blob/a9b42574d9f608c241ee92995a2d73d0926965de/src/api/catalog-search/catalog-search.ts#L143 This information gets propagated to packages so that some tags will be listed under both "tags" field and the "keywords" field.

- The second source of duplicate keywords is from keywords provided by authors naturally overlapping with package tags we have added.   e.g. We add a "Kubernetes" tag, and a library author has a "kubernetes" keyword.

- The third source of duplicate keywords is that some `packageTags` we specify on our ConstructHub instance are assigned the same keyword label.

The easiest fix without reverting #694 is to de-duplicate tags at rendering time, always preferring the tag with a given label before a keyword with the same name. This solution ensures that the same order of tags/keywords is shown on both on search results and on package pages.


eladb (commented):
Test?

gabewomble (approved):

PR #704, opened 2021-11-18
fix: api ref nav behaves differently from readme

eladb:
When clicking on a nav item from readme, it will scroll the view so that this item is at the top (which implies that the package header is not displayed). When clicking on an API reference item, it will scroll to 0 (which implies that package header is displayed). This results in an inconsistent exprience.

To solve this, we added an `id` to the API doc header and scroll it to view if the URL is an `/api` URL.

Additionally, we also made this header a link, so that it will behave like README headers (which are links to themselves).

### DIFF
--- a/src/views/Package/PackageDocs.tsx
+++ b/src/views/Package/PackageDocs.tsx
@@ -46,7 +46,6 @@ export const PackageDocs: FunctionComponent = () => {
       target?.scrollIntoView(true);
     } else if (isApiPath(pathname)) {
       const target = document.getElementById(DOCS_ROOT_ID) as HTMLElement;
-      console.log(target);
       target?.scrollIntoView(true);
     } else {
       window.scrollTo(0, 0);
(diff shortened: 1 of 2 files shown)


src/views/Package/PackageDocs.tsx
```diff
@@ -26,15 +28,26 @@ const SubmoduleSelector: FunctionComponent = () => {
...
-  const { hash } = useLocation();
+  const { hash, pathname } = useLocation();
+
   useEffect(() => {
     if (hash) {
       const target = document.querySelector(`${hash}`) as HTMLElement;
       target?.scrollIntoView(true);
+    } else if (isApiPath(pathname)) {
+      const target = document.getElementById(DOCS_ROOT_ID) as HTMLElement;
+      console.log(target);
```
gabewomble, line 49: ✂️

eladb: done

gabewomble (approved):

PR #706, opened 2021-11-18
chore: improve assembly type safety

eladb:
Since backend now trims down some fields in `assembly.json`, it would be wise to reflect that in the downstream types so that client code won't accidentally use these fields in the future.

### DIFF
--- a/src/api/package/assembly.ts
+++ b/src/api/package/assembly.ts
@@ -27,8 +27,5 @@ export const fetchAssembly = async (
       `Failed fetching assembly for ${assemblyPath}: ${response.statusText}`
     );
   }
-  const json = (await response.json()) as spec.Assembly;
-  delete json.types;
-  delete json.readme;
-  return json;
+  return response.json();
 };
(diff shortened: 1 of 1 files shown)


src/api/package/assembly.ts
```diff
@@ -2,14 +2,22 @@ import * as spec from "@jsii/spec";
 import { API_PATHS } from "../../constants/url";
 import { getAssetsPath } from "./util";
 
+// These fields are removed from assembly.json during processing to save space,
+// and we don't need them here in the client.
+// See https://github.com/cdklabs/construct-hub/pull/567
+export type SlimAssembly = Omit<
+  spec.Assembly,
+  "types" | "readme" | "dependencyClosure"
```
iliapolo, line 10: Where is the `readme` and `dependencyClosure` removed? I only see `types` being removed.

https://github.com/cdklabs/construct-hub/blob/main/src/backend/ingestion/ingestion.lambda.ts#L106

eladb: Still haven't gotten around to it, but I am on it.

gabewomble (approved):

src/api/package/assembly.ts
eladb: https://github.com/cdklabs/construct-hub/pull/573

PR #709, opened 2021-11-18
feat: final home page copy and sections

### LINKED ISSUE #647: Implement the final version of the home page messages
Requires discussion with Song-I

gabewomble:
Fixes #647 

image


Chriscbr (changes requested):
Awesome! These copy improvements look great, as well as the new layout of sections. 🙂 

Since our assets' names don't change between builds and might be cached by the end-user's browser, I think we should avoid modifying them in-place to eliminate the possibility of a client rendering the page with cached versions after we deploy this change.  Maybe we could add a "-v2" to the names just to play things safe? What do you think?

Chriscbr (approved):

PR #723, opened 2021-11-19
fix: language bar links to invalid URLs

### LINKED ISSUE #715: Change the behavior on a language change
Currently, if you go to this link 
https://constructs.dev/packages/constructs/v/3.3.161/api/constructsconstruct?lang=python

and change the language to Java, you get an error page.

We should make sure than whenever the link includes a reference to a header in the documentation (in this case /api/constructsconstruct and the user presses another programming language it will reload the entire page, means: the link will be https://constructs.dev/packages/constructs/v/3.3.161/?lang=java and not https://constructs.dev/packages/constructs/v/3.3.161/api/constructsconstruct?lang=java

Chriscbr:
Fixes #715

### DIFF
--- a/src/components/LanguageBar/LanguageBar.tsx
+++ b/src/components/LanguageBar/LanguageBar.tsx
@@ -1,11 +1,14 @@
 import { IconButton, Stack } from "@chakra-ui/react";
 import type { FunctionComponent } from "react";
+import { useHistory } from "react-router-dom";
 import {
   Language,
   TEMP_SUPPORTED_LANGUAGES,
   LANGUAGE_RENDER_MAP,
   LANGUAGES,
 } from "../../constants/languages";
+import { getPackagePath } from "../../util/url";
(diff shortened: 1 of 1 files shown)


src/components/LanguageBar/LanguageBar.tsx
```diff
@@ -17,8 +20,9 @@ export interface LanguageBarProps {
 export const LanguageBar: FunctionComponent<LanguageBarProps> = ({
   targetLanguages,
   selectedLanguage,
-  setSelectedLanguage,
 }) => {
+  const { name, version } = usePackageState();
```
Chriscbr, line 24: I wasn't sure if there's any advantage/disadvantage retrieving state here or passing package state from the parent component (`LanguageSelection`). LMK if either is better

gabewomble: I think this is fine 👍

gabewomble (approved):

PR #738, opened 2021-11-22
chore: add home page fallback if catalog.json doesn't exist

Chriscbr:
Adds an extra fallback mechanism so that users will still be able to see the home page (without any of the "package" sections) in the event that `catalog.json` is unavailable. Currently we throw an error when this happens, which would cause the user to see a blank page.

Related to https://github.com/cdklabs/construct-hub-internal/issues/74


src/views/Home/Home.tsx
```diff
@@ -35,11 +39,11 @@ export const Home: FunctionComponent = () => {
...
         <Info />
 
-        {fullSite ? <Categories /> : <></>}
+        {hasPackages && fullSite ? <Categories /> : <></>}
 
-        {fullSite ? <CDKTypeTabs /> : <></>}
+        {hasPackages && fullSite ? <CDKTypeTabs /> : <></>}
 
-        {fullSite ? <Featured /> : <></>}
+        {hasPackages && fullSite ? <Featured /> : <></>}
```
gabewomble, line 46: For some reason it won't let me make a suggestion here, but can we clean this up to be

```tsx
{hasPackages && fullSite && (
  <>
    <Categories />
    <CDKTypeTabs />
    <Featured />
  </>
)}
```

Chriscbr: Done

gabewomble (approved):

addihorowitz (approved):

PR #740, opened 2021-11-22
fix: wrong number of constructs in search bar

### LINKED ISSUE #739: Total number of constructs - should be fixed
https://constructs.dev/search?q=&offset=0 says we have 795 packages.
Search bar says 900+.

Chriscbr:
Fixes #739

### DIFF
--- a/src/components/SearchBar/SearchBar.tsx
+++ b/src/components/SearchBar/SearchBar.tsx
@@ -18,8 +18,8 @@ import {
   ChangeEventHandler,
   FormEventHandler,
 } from "react";
-import { useCatalog } from "../../contexts/Catalog";
 import { useCatalogSearch } from "../../hooks/useCatalogSearch";
+import { useSearch } from "../../hooks/useSearch";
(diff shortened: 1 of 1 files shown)


iliapolo:
@Chriscbr Can you explain the root cause of the problem? Its not clear, at least to me, from the code.

iliapolo:
Also don't forget to cherry pick to `main`, or alternatively reset the base of this PR to `main` and cherry-pick to `dev`

MrArnoldPalmer (approved):

Chriscbr:
@iliapolo Yep. The main issue is that this component was calculating the total number of packages based on the raw catalog data, which has distinct items for each major version. Instead, it should use one of the higher-level APIs/hooks that is integrated with our catalog search API, which automatically dedups packages.

My mental model right now is that our hooks build on top of each other like this:
`useCatalog (raw data) <-- useSearch (adds dedup, query, filter, sort, etc.) <-- useCatalogResults (adds pagination)`

PR #757, opened 2021-11-23
chore: trigger ch main upgrade on release

iliapolo:
Now that CH `main` upgrade workflow [only upgrades the webapp](https://github.com/cdklabs/construct-hub/pull/607), its ok for a webapp release to trigger it.


eladb (approved):

eladb:
You'd want to also push this to `main`

PR #792, opened 2021-11-29
fix: improve keyword searching

### LINKED ISSUE #789: A search for "dyanmodb" retrieves 23 results, while searching by keywords retrieves more results
https://constructs.dev/search?q=dynamodb&offset=0

https://constructs.dev/search?q=&keywords=amazon+dynamodb&offset=0

It seems like the search algorithm by search terms is still not running on the keywords.

gabewomble:
Fixes #789 

This change flattens keyword tokens so that when we search by keywords, we can find results for partial matches

Before, the query for `dynamodb` would _only_ match the `dynamodb` keyword, whereas now it will also match `amazon dynamodb`

Changes outside of the `catalog-search` file are all for lint updates

### DIFF
--- a/src/api/catalog-search/catalog-search.test.ts
+++ b/src/api/catalog-search/catalog-search.test.ts
@@ -78,6 +78,31 @@ describe("CatalogSearchAPI", () => {
     );
   });
 
+  it("Support partially matched keywords", () => {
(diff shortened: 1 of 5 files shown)


Chriscbr (changes requested):
Can we add a unit test for this?

Chriscbr (approved):

PR #800, opened 2021-12-01
fix: support nav items which do not start with letters, reimplement active nav item state

### LINKED ISSUE #736: document sidebar with hash prefix will be ignored
image

image

gabewomble:
Fixes #736 

Since we cannot guarantee that markdown headers will start with a letter, we can't query select markdown headings by their id. Instead, we can use the `data-heading-id` attribute and maintain the same functionality while preventing an error from querying invalid id's.

Additionally, this work fixes the Nav Item active states so that the active nav item will be colored blue - which was prior behavior to the markdown rendering changes from @MrArnoldPalmer

To test this work, I had to download a markdown response for a package, and copy it to my public folder with the corresponding request path (`localhost:3000/data/foo/v123/docs-typescript.md` => `public/data/foo/v123/docs-typescript.md`). This allowed me to modify the markdown headings freely for testing as the original library with this issue no longer has numbered headings.

Screen Shot 2021-12-01 at 11 09 32 AM

### DIFF
--- a/src/components/Markdown/Headings.tsx
+++ b/src/components/Markdown/Headings.tsx
@@ -68,7 +68,7 @@ export const Headings: FunctionComponent<HeadingResolverProps> = ({
         size={size}
       >
         <NavLink
-          data-heading-id={id}
+          data-heading-id={`#${id}`}
           data-heading-level={level}
           data-heading-title={title}
           id={id}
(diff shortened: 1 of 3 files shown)


Chriscbr (approved):
> Additionally, this work fixes the Nav Item active states so that the active nav item will be colored blue

sweet!

PR #804, opened 2021-12-05
chore: upgrade deps

eladb:
Signed-off-by: Elad Ben-Israel <benisrae@amazon.com>


rix0rrr (approved):

gabewomble (changes requested):
This version of eslint isn't compatible with `react-app-rewired`, meaning that it isn't possible to run a dev server or build the app with these changes

addihorowitz (approved):

MrArnoldPalmer:
> We have a bigger problem with the react tool chain. It appears they are unwilling to upgrade vulnerable dependencies.

I'm on the fence as to whether this is actually a problem for us. As projen and react-scripts are both build time tools, if these vulnerabilities are in fact not relevant and it creates a significant amount of noise/work for the maintainers, letting them lie I think is fine.

>We will either need to convince them to maintain their dependencies or we will have to eject from this tool and reimplement it ourselves.

What are our arguments to convince them that haven't already been presented and addressed? If we agree that the remaining vulnerabilities cannot be exploited, it's basically "we don't want npm audit or dependabot to complain", to which we can get a clean run by moving everything to `devDependencies` (which is accurate for our project) and running `npm audit --production` and by dismissing the alerts on github as "vulnerable code is not actually used". If we don't agree that they are irrelevant then we need to dig up that evidence/usecase and present it.

That being said, I think we should just dump react-scripts from the react projen template and instead output a very basic webpack configuration that users can add to/override via `.projenrc.js`. After using CRA with projen in construct hub my opinion is that the two are essentially in conflict and create a lot of pain points for the user when they want to break out of the CRA defaults. Basically in CRA, you can `eject` and take over, but with projen + CRA, projen has no concept of what the ejected artifacts are and how to configure and maintain them, leading to needing `react-app-rewired` etc.

For construct-hub-webapp's usecase, it would be a fairly simple webpack configuration needed to build our app, and one that @gabewomble and I could probably implement fairly easily. However, with our desire to move to NextJS we should weigh effort vs long term benefits.

TLDR: I'm not worried about these vulnerabilities right now for construct-hub-webapp. `react-scripts` is of course used only at build time and the published artifact on npm is just a bunch of static webapp files with no runtime dependencies. We should update our dependency list to reflect that, nothing should be a runtime dep. I trust that the maintainers of `react-scripts` have thoroughly examined the reported vulnerabilities and found them irrelevant/impossible to exploit.

But also let's just rewrite the react-app template to not use CRA anyway cause it works better for projen users.

PR #817, opened 2021-12-15
chore: optimizing search state

gabewomble:
This pull request introduces [RecoilJS](https://recoiljs.org/) as a solution for global state management.

This allows us to atomize individual pieces of search state so that components only re-render when the data which they depend on changes, rather than when the context they subscribe to changes.

How search re-rendered before:

https://user-images.githubusercontent.com/8749859/146257967-46b244d6-05a4-4e52-bd9d-8b3587a0c437.mov

How search re-renders on this branch:

https://user-images.githubusercontent.com/8749859/146258134-487f2115-8ae9-4072-81d3-a8319517ea49.mov


src/state/search.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export const pageState = atom<ExtendedCatalogPackage[]>({
+  key: "search-pageState",
+  default: [],
+});
+
+export const tagsState = atom<string[]>({
+  key: "search-tagsState",
+  default: [],
+});
```
MrArnoldPalmer, line 63: As discussed briefly separately, most of these are duplicated from the url query parameters. In something like redux in the past, I would have argued against manually synchronizing these to the store and instead pulling them from the URL, and dispatching actions to change the URL directly to ensure a single source of truth and a one way data flow.

That being said, is it possible to only have a single atom for `resultsState` that is the data from catalog.json, then the selector takes that and the query params as input to compute the filtered results to be displayed, and components dispatch changes directly to the query params that they need to in order to change results?

src/index.tsx
```diff
@@ -21,27 +22,29 @@ const queryClient = new QueryClient();
...
-      </Router>
+      <RecoilRoot>
+        <Router>
+          <ShortbreadProvider>
+            <AnalyticsProvider>
+              <ConfigProvider>
+                <CatalogProvider>
+                  <StatsProvider>
+                    <VersionsProvider>
+                      <SearchProvider>
```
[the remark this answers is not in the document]
gabewomble: Not any longer. In a follow up I'd like to consolidate our context providers into atoms as much as possible

src/state/search.ts
gabewomble: This might make the most sense to do as a follow-up item. I plan on migrating our data providers to atoms rather than react-contexts, which would pave the way to the work you're suggesting.

gabewomble:
Closing as I've found a simpler solution that does not require juggling global vs url state

PR #820, opened 2021-12-17
feat: e2e canary tests

Chriscbr:
Fixes https://github.com/cdklabs/construct-hub/issues/631

See comment in .projenrc.js for additional details.

I set the canary to run once every three hours, so that we don't necessarily get inundated with issues if the canary starts failing.


.projenrc.js
```diff
@@ -157,6 +157,79 @@ const project = new web.ReactTypeScriptProject({
...
+        },
+        {
+          name: "Upload videos",
+          uses: "actions/upload-artifact@v2",
+          if: "always()",
+          with: {
+            name: "cypress-videos",
+            path: "cypress/videos",
+          },
+        },
```
gabewomble, line 216: It seems like these `steps` are the same that we use on the cypress PR workflow. I think it would make sense to consolidate the two into a single `cypressSteps` array

gabewomble (approved):
Awesome!

PR #827, opened 2021-12-26
chore(deps): upgrade dependencies

iliapolo:
Replaces https://github.com/cdklabs/construct-hub-webapp/pull/826 since that latest projen version requires some code changes.


.projenrc.js
```diff
@@ -26,8 +26,7 @@ const project = new web.ReactTypeScriptProject({
     },
   },
 
-  minNodeVersion: "12.20.0",
-  workflowNodeVersion: "12.22.0",
+  minNodeVersion: "16.0.0",
```
iliapolo, line 29: `react-scripts` requires node 14.0.0, but `eslint` requires:

```console
eslint@8.5.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "14.15.4"
```

Figured we just bump to 16 and be done with it.

.projenrc.js
```diff
@@ -148,16 +147,17 @@ const project = new web.ReactTypeScriptProject({
...
-    cypress: {
-      name: "E2E Tests",
-      runsOn: "ubuntu-latest",
-      permissions: {
-        checks: "write",
-        contents: "read",
-      },
-      steps: cypressRunSteps,
+
+  const integWorkflow = project.github.addWorkflow("integ");
```
iliapolo, line 151: The latest version of projen doesn't offer an `addJob` method on `buildWorkflow`, instead it offers `addPostBuildJob`, which is intended for jobs that requires the build artifact. 

These tests don't actually need anything from the build job, so decided to move it to a separate workflow. Similarly to how we do it in [projen](https://github.com/projen/projen/blob/main/.github/workflows/integ.yml) itself.

.projenrc.js
```diff
@@ -148,16 +147,17 @@ const project = new web.ReactTypeScriptProject({
...
+  integWorkflow.on({ workflowDispatch: {}, pullRequest: {} });
+  integWorkflow.addJob("e2e", {
+    name: "e2e",
```
iliapolo, line 154: my OCD to keep job names lower cased consistent.

eladb (approved):

PR #829, opened 2021-12-26
chore(deps): resolve `nth-check` CVE

iliapolo:
Had to forcefully use `resolutions` because `react-scripts` currently depends (eventually) on the bad version. 

Address https://github.com/cdklabs/construct-hub-webapp/security/dependabot/yarn.lock/nth-check/open


eladb (approved):

.projenrc.js
```diff
@@ -88,6 +88,11 @@ const project = new web.ReactTypeScriptProject({
   autoApproveUpgrades: true,
 });
 
+// addressing https://github.com/advisories/GHSA-rp65-9cf3-cjxr forcefully until
+// react-scripts fixes the dependency chain.
+// if test pass, we should be ok with this override, even though its a different major version.
+project.package.addField("resolutions", { "nth-check": "2.0.1" });
```
eladb, line 94: Would be nice if we could do:

```ts
project.package.addResolutions('nth-check', '2.0.1');
```

And it will only work for yarn (because I think npm doesn't support it)

PR #831, opened 2021-12-26
chore(deps): downgrade to node 14

iliapolo:
I was a bit too trigger happy with https://github.com/cdklabs/construct-hub-webapp/pull/827#discussion_r775239999. 

This will force `construct-hub` to declare node 16 as well, but its not yet available as a lambda runtime, and we don't want to be developing on a node version higher than what we are deploying on.


iliapolo:
Using my admin rights to merge this - confident enough the PR is ok.

PR #833, opened 2021-12-28
feat: support packageTagGroups config

gabewomble:
Related to: https://github.com/cdklabs/construct-hub/pull/685

This front end feature work enables support for an unreleased `packageTagGroups` config option, allowing construct hub operators to customize the label, tooltip, and filter type for search tag groups. The changes are backwards compatible, so there should be no impact to existing functionality

Here is a picture from my local development, which proxies production

Screen Shot 2021-12-27 at 4 10 15 PM

And here is a screenshot from a custom deployment using the new backend configuration. Available at https://d3u0cndfkrxv5v.cloudfront.net/search?q=&offset=0

Note that the tags themselves haven't been applied to the catalog yet

Screen Shot 2021-12-27 at 4 10 52 PM


src/api/config/index.ts
```diff
@@ -25,6 +25,13 @@ export interface PackageTagConfig {
   };
 }
 
+export interface TagGroupConfig {
```
Chriscbr, line 28: Can we add a small comment in the source code that links/references the original interface defined in our backend repo? May help future readers

src/views/Search/TagFilter.tsx
```diff
@@ -1,31 +1,76 @@
...
+      map.set(group.id, group);
+    });
+
+    return map;
+  }, [packageTagGroups]);
+
+  const tagFilterGroups: FilterGroups = useMemo(
+    () =>
+      filterableTags.reduce(
+        (accum: FilterGroups, tag: PackageTagConfig): FilterGroups => {
```
Chriscbr, line 38: This useMemo function has a bit of complexity to it - can extract part of this into a utility function and unit test it perhaps?

Chriscbr (changes requested):
Looks good! Just two small comments 🙂

src/views/Search/TagFilter.tsx
gabewomble: Good idea 👍

Chriscbr (approved):

Issue #824, opened 2021-12-21
Package submodule API reference does not load


Chriscbr:
If you visit https://constructs.dev/packages/monocdk-experiment/v/1.66.0, click on "submodules", and then choose a module, the API table of contents does not update and no new content is displayed (though the URL is updated to add the "submodule=xxx" query parameter).

But if you refresh the page, then the correct table of contents is updated.

(optional) - When a submodule is selected but no API has been chosen, the content of the page should probably not be blank -- perhaps it could just display the text "Please choose an API from the table of contents".

PR #852, opened 2022-01-04
fix: update docs when submodule changes

gabewomble:
Fixes https://github.com/cdklabs/construct-hub-webapp/issues/824

This PR primarily fixes the above bug while also handling some related issues:
1. Update doc fetching hook to listen for submodule changes (solves root issue)
2. Add error message to display for modules or submodules that do not have a readme
3. Update url handlers for submodule selection to use the tested `getPackagePath` url helper (removes unwanted query params)

### DIFF
--- a/src/views/Package/ChooseSubmodule/ChooseSubmodule.tsx
+++ b/src/views/Package/ChooseSubmodule/ChooseSubmodule.tsx
@@ -1,24 +1,19 @@
 import { ArrowBackIcon, ChevronDownIcon } from "@chakra-ui/icons";
 import { Button, Divider, Stack, useDisclosure } from "@chakra-ui/react";
-import type { Assembly } from "@jsii/spec";
 import { FunctionComponent, useCallback, useMemo, useState } from "react";
-import { useHistory, useLocation } from "react-router-dom";
+import { useHistory } from "react-router-dom";
 import { QUERY_PARAMS } from "../../../constants/url";
 import { useQueryParams } from "../../../hooks/useQueryParams";
+import { getPackagePath } from "../../../util/url";
+import { usePackageState } from "../PackageState";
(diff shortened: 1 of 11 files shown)


src/views/Package/ChooseSubmodule/ChooseSubmodule.tsx
```diff
@@ -28,18 +23,28 @@ export const ChooseSubmodule: FunctionComponent<ChooseSubmoduleProps> = ({
...
   const onGoBack = () => {
-    const lang = query.get(QUERY_PARAMS.LANGUAGE);
-    push(`${pathname}${lang ? `?${QUERY_PARAMS.LANGUAGE}=${lang}` : ""}`);
+    push(
+      getPackagePath({
+        name: packageName,
+        version,
+        language,
+      })
+    );
```
gabewomble, line 35: Previously we were preserving query params and hashes which caused unwanted behavior. If a user was viewing the docs for Construct Foo within submodule `Bar` and then switched docs to a different submodule or the main module, the page would try to render docs from the same path (Construct Foo), which would not exist

src/views/Package/PackageState.tsx
```diff
@@ -70,7 +70,7 @@ export const PackageStateProvider: FunctionComponent = ({ children }) => {
   const pkgName = scope ? `${scope}/${name}` : name;
   const allVersions = versionData?.packages[pkgName];
 
-  const markdown = useQuery(`${id}-docs-${language}`, () =>
+  const markdown = useQuery(`${id}-docs-${language}-${submodule}`, () =>
```
gabewomble, line 73: All that was necessary to fix the core issue was updating this request id to take a dependency on the submodule

MrArnoldPalmer (approved):

PR #884, opened 2022-01-28
chore: update TOC to be fixed width

Chriscbr:
Updates the table of contents so that it always has the same width on desktop viewports, as it can seem overly large or overly squished on certain screen resolutions.

Also updates the table of contents to display ellipses when titles get too long -- the CSS for ellipses was already there, but the text was still wrapping since the `white-space` property wasn't set to `nowrap`.

Screen Shot 2022-01-27 at 4 16 08 PM

### DIFF
--- a/src/views/Package/PackageDocs.tsx
+++ b/src/views/Package/PackageDocs.tsx
@@ -67,7 +67,7 @@ export const PackageDocs: FunctionComponent = () => {
     <Grid
       columnGap={4}
       h="full"
-      templateColumns={{ base: "1fr", md: "20rem 3fr" }}
+      templateColumns={{ base: "1fr", md: "minmax(20rem, 1fr) 3fr" }}
       width="100%"
     >
       <Flex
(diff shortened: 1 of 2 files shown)


src/views/Package/PackageDocs.tsx
```diff
@@ -67,7 +67,7 @@ export const PackageDocs: FunctionComponent = () => {
     <Grid
       columnGap={4}
       h="full"
-      templateColumns={{ base: "1fr", md: "1fr 3fr" }}
+      templateColumns={{ base: "1fr", md: "20rem 3fr" }}
```
gabewomble, line 70: Could you try using `minmax(20rem, 1fr)` instead? I love the addition of setting a minimum width, but we can leverage the `minmax` function so that users with wider resolutions aren't constrained to the minimum `20rem` size

gabewomble (changes requested):
One small nit, but otherwise this is a great addition!

src/views/Package/PackageDocs.tsx
Chriscbr: Sure thing!

gabewomble (approved):

PR #885, opened 2022-01-28
feat: add random package route

Chriscbr:
Visit constructs.dev/random (or your-construct-hub.com/random) to get taken to a random construct.

This is something I thought would be useful for testing purposes so that I can easily visit a random package to see how the API reference looks, whether tags are appearing properly, etc. By contrast if I find packages through search, I tend to be biased towards packages that are recently published, or are ranked highest/lowest in downloads, or come at the beginning/end of the alphabet, etc.

For now this is just an easter egg basically, but I'm totally open to adding it to the UI. 🙂


gabewomble (approved):
I love this just the way it is 🔥 🔥 🔥

PR #893, opened 2022-02-01
fix: links within API reference take longer to load

### LINKED ISSUE #892: Links within docs take longer to load than links within table of contents
Example: if you are viewing the ConstructHub construct (http://constructs.dev/packages/construct-hub/v/0.3.222/api/ConstructHub?lang=typescript) and click on the link to ConstructHubProps in the table of contents, the new content will update in-place without having reload the entire page.

However, if you go back to the ConstructHub page and instead click on the link to ConstructHubProps found within the "Initializers" section, the entire page gets reloaded. This is jarring if you're using dark mode since the page momentarily flashes white for a fraction of a second as new page state is loaded/rendered.

Chriscbr:
Fixes #892 by changing all non-external links to implicitly get treated as internal navigation links.

### DIFF
--- a/src/components/Markdown/Text.tsx
+++ b/src/components/Markdown/Text.tsx
@@ -1,41 +1,41 @@
-import { Box, HTMLChakraProps, Link, Text } from "@chakra-ui/react";
+import { Box, PropsOf, Text } from "@chakra-ui/react";
 import { Language } from "prism-react-renderer";
 import type { FunctionComponent } from "react";
 import { ExternalLink } from "../ExternalLink";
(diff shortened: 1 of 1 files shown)


Chriscbr:
@gabewomble I wonder if this approach makes sense... could there be cases where the URL in the markdown isn't a valid external link, but isn't necessarily internal either?

gabewomble (approved):
@Chriscbr I think this approach makes sense / works fine. I don't know of any occurrences of the scenario you mentioned, but if it happens our app should 404, which is ok

PR #894, opened 2022-02-01
feat: separate link from markdown heading

### LINKED ISSUE #867: Separate anchor link in API reference headers
Currently, all headers on API reference pages that can be linked to are direct links. This is a small pain point since it makes it harder to copy the names of headers, which is common since they are often the names of methods and constructs that you want to use in your code.

I wonder if it might be an improvement to change the headers so they no longer link, and add a separate ornament that appears on mouseover, which either takes links to the correct anchor, or which saves a link to your clipboard to the appropriate API reference item.

Examples of this UI pattern in other docs websites:

----

(pkg.go.dev)
Screen Shot 2022-01-13 at 2 56 03 PM

----

(stripe.com)
Screen Shot 2022-01-13 at 2 58 09 PM

----

(developer.android.com)
Screen Shot 2022-01-13 at 3 00 55 PM

gabewomble:
Fixes #867 

Separates the anchor element from the markdown heading to make it easier to copy headings.
When hovering the header line, the link icon will become visible. It will remain visible if clicked / focused

Screen Shot 2022-02-01 at 1 19 56 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ jspm_packages/
 !/.eslintrc.json
 /build/
 !/react-app-env.d.ts
+/.vscode/
 cypress/videos/
 cypress/screenshots/
 !/.github/workflows/integ.yml
--- a/.npmignore
+++ b/.npmignore
@@ -12,12 +12,12 @@
 dist
 /tsconfig.json
 /.github/
-/.vscode/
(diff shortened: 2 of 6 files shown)


Chriscbr (approved):
LGTM!

PR #895, opened 2022-02-01
feat: class navigation sidebar

### LINKED ISSUE #319: Add list of class properties and methods for quick navigation
I frequently look up class properties and methods when working with constructs. Having the list on the right side in AWS CDK API Reference allows me to quickly understand what is available. This is very important because I don't want to read the whole documentation page to find what I need, a property or method name is often enough. Then I can quickly jump to the documentation of that property or method.

AWS CDK API Reference | Construct Hub
-- | --
image | image

Chriscbr:
Closes #319

Adds a class navigation sidebar for quickly navigating methods, properties, etc.

To avoid too much squishing because of the extra horizontal room it takes up, I changed the responsive layout for package pages so that the switching point between tablet and desktop layouts is at chakra's "lg" instead of "md".

Currently this sidebar is created based on the JSON doc schema that is generated in conjunction with markdown docs for each package. Since these are different sources of information, there's some logic that had to be duplicated from how we are generating the markdown and generating links / IDs. Once we generate the API reference straight from the API.json and remove the markdown from the picture, it's possible we could simplify this down a lot.

Screenshot:

Screen Shot 2022-02-01 at 3 33 55 PM


src/api/package/docs.ts
```diff
@@ -35,3 +36,43 @@ export const fetchMarkdown = async (
...
+  const expectedContentType = "application/json";
+  const contentType = response.headers.get("Content-Type");
+
+  // we check 'includes' and not 'equal' because the content type contains
+  // charset as well (e.g text/markdown; charset=UTF-8)
+  if (!contentType || !contentType.includes(expectedContentType)) {
+    throw new Error(
+      `Invalid content type: ${contentType}. Expected ${expectedContentType}"`
+    );
+  }
```
Chriscbr, line 75: The comments here are copied from the similar `fetchMarkdownDocs` function above.

Note - this is currently raising errors because of the bug caught in https://github.com/cdklabs/construct-hub/pull/740 -- once our data on constructs.dev gets reprocessed this should be fixed, but I'll make sure to validate this before merging

src/api/package/docs.ts
```diff
@@ -35,3 +36,42 @@ export const fetchMarkdown = async (
...
+ * Fetch JSON docs of a specific package from the backend.
+ */
+export const fetchJsonDocs = async (
+  name: string,
+  version: string,
+  language: string,
+  scope?: string,
+  submodule?: string
+): Promise<Schema> => {
+  const docsSuffix = `/docs-${
```
gabewomble, line 50: Nit: could you generalize this to a local `getDocsSuffix` helper so we can re-use it in both `fetchMarkdownDocs` and `fetchJsonDocs`?

src/components/NavTree/NavTree.tsx
```diff
@@ -19,6 +19,7 @@ export interface NavItemProps extends NavItemConfig {
   // The following props don't need to be explicitly defined - they are passed internally
   onOpen?: () => void;
   level: number;
+  styleProps?: any;
```
gabewomble, line 22: Exposing a `styleProps` prop is inconsistent with how Chakra handles styles. To avoid over-complicating an alternative solution, I think it would make a lot of sense to expose a `variant` prop, which would set the necessary styles internally

You can also reference the [chakra component styling doc](https://chakra-ui.com/docs/theming/component-style) if you want to implement this the "chakra way"

src/views/Package/PackageDocs.tsx
```diff
@@ -42,9 +49,126 @@ const isApiPath = (path: string) => {
...
+ * in the JSON docs play well with the existing markdown docs, but this could be
+ * refactored / dropped in the future.
+ */
+const docsSectionsMap = {
+  initializer: {
+    header: "Initializer Props",
+    getValues: (e: ClassSchema) => e.initializer?.parameters ?? [],
+    // @aws-cdk/aws-s3.Bucket.Initializer.parameter.scope => Bucket#Initializer.parameter.scope
+    getPath: (id: string) =>
+      id.split(".").slice(-4).join(".").replace(".", "#"),
```
gabewomble, line 64: Can we generalize this `getPath` helper? Looks like the only difference between these functions is the slice index

src/views/Package/PackageDocs.tsx
```diff
@@ -114,6 +241,27 @@ export const PackageDocs: FunctionComponent = () => {
...
+        borderLeftColor="borderColor"
+        direction="column"
+        display={{ base: "none", lg: "flex" }}
+        maxHeight={`calc(100vh - ${TOP_OFFSET})`}
+        overflow="hidden auto"
+        pl={6}
+        position="sticky"
+        pr={4}
+        top={TOP_OFFSET}
+      >
```
gabewomble, line 256: This flex container has a lot in common with the main nav container. We can create a common component for the two of them either in a separate file:

```tsx
// Feel free to name it something catchier
export const StickyNavContainer: typeof Flex = (props) => (
  <Flex
    as="nav" // to be more semantic
    alignSelf="stretch"
    direction="column"
    display={{ base: "none", lg: "flex" }}
    maxHeight={`calc(100vh - ${TOP_OFFSET})`}
    overflow="hidden auto"
    position="sticky"
    {...props}
  />
);

```

gabewomble (changes requested):
This is really solid work! Left a few comments I'd like to see resolved, but I can't wait to see this killer feature in action 😎

src/views/Package/PackageDocs.tsx
Chriscbr: Done

src/components/NavTree/NavTree.tsx
Chriscbr: Gotcha, yeah I wasn't too sure about this. I've switched it out with a "variant" field 👍

Chriscbr:
Just noticed one more small issue, which is that the query params aren't properly added to the links - fixing...

src/views/Package/PackageDocs.tsx
Chriscbr: Updated

gabewomble (approved):
Awesome work 🤘

PR #897, opened 2022-02-02
fix: md links for submodules

gabewomble:
When browsing docs for a submodule, clicking on a heading link does not work as expected. A long-term fix to this would be to properly route submodules rather than treating them as a query param, but for now we just need to persist the query param within the hash link's href to avoid breaking the experience.

### DIFF
--- a/src/components/Markdown/Headings.tsx
+++ b/src/components/Markdown/Headings.tsx
@@ -1,4 +1,5 @@
-import { Heading, As } from "@chakra-ui/react";
+import { LinkIcon } from "@chakra-ui/icons";
+import { Flex, Heading, As } from "@chakra-ui/react";
 import { Children, FunctionComponent, ReactNode } from "react";
 import ReactDOMServer from "react-dom/server";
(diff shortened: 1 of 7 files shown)


src/components/Markdown/Headings.tsx
```diff
@@ -35,25 +36,29 @@ const HeadingLink: FunctionComponent<{
...
+      alignItems="center"
+      data-heading-id={`#${id}`}
+      data-heading-level={level}
+      data-heading-title={title}
+      display="flex"
+      id={id}
+      lineHeight={1}
+      opacity="hidden"
+      replace
+      to={`${search}#${id}`}
```
Chriscbr, line 55: Can we add a comment explaining what this means / what an example of this value `${search}#${id}` would be?

Chriscbr (approved):
Looks good to me - I added the do-not-merge label so you can address the one comment.

src/components/Markdown/Headings.tsx
gabewomble: Sure 👍

PR #912, opened 2022-02-14
feat: support multiple cdk types on packages

### LINKED ISSUE #900: Support multiple CDK types in package metadata
Following the discussion on this issue - https://github.com/cdklabs/construct-hub/issues/717

We have discovered a use case for multiple cdk types, so we need to support the functionality! This task should support receiving the CDK type metadata as an array, and for backwards compatibility, map the existing metadata to an array.

Acceptance Criteria:
1. Webapp should support constructs with multiple cdk types listed in metadata
2. Constructs with multiple cdk types should be visible in search results filtered by any cdk type the construct is associated with. (E.g library foo has cdk8s and cdktf, so it should show in the results for cdk8s or cdktf)
3. Constructs with multiple CDK types should have a badge that reads "Multi-CDK". When hovered, they should display a tooltip that enumerates the CDK types and versions supported

gabewomble:
Fixes #900 

Implements backward compatible support for multi cdk type constructs in construct hub. When sorting by CDK type, "Multi-CDK" constructs will show in the results for any of their matching CDK types. Additionally, they will have a "Multi-CDK" badge, which shows a tooltip when hovered.

The tooltip (now) has an arrow and is placed to the left

Screen Shot 2022-02-14 at 1 29 21 PM

Screen Shot 2022-02-14 at 2 40 23 PM


src/api/package/metadata.ts
```diff
@@ -3,11 +3,24 @@ import { API_PATHS } from "../../constants/url";
...
-  constructFramework?: {
-    name?: CDKType;
-    majorVersion?: number;
-  };
+  /**
+   * Describes the associated Construct framework for a library.
+   * Back-end will introduce a new metadata.constructFrameworks property to replace metadata.constructFramework
+   * @deprecated Use constructFrameworks instead
+   */
+  constructFramework?: Partial<ConstructFramework>;
```
Chriscbr, line 17: Why is this partial?

Chriscbr (approved):
Great work! 😄

src/api/package/metadata.ts
gabewomble: Just so that I could re-use the type for the new implementation. The old signature allowed `cdkType` and `majorVersion` to be undefined, whereas moving forward we should only define a constructFramework object if there is a `cdkType` associated

PR #915, opened 2022-02-15
Feat/move readme subitems to right nav

### LINKED ISSUE #914: Display README sub-items on right-hand nav
To be more consistent with the API Reference doc UX, we should move readme sub-headings to the right hand navigation menu

gabewomble:
Fixes #914 

This pull request moves README subitems from the primary doc navigation to the secondary nav. The first commit on this pull request re-organizes logical code from presentational code, while the second commit actually does the work to implement the feature.

While testing this feature, I've also observed that libraries which are missing readmes (common for submodules) have an unclear UX and we should try to add some messaging to clarify that scenario i.e "This {library or submodule} does not have a readme. Use the menu on the left to browse the API Reference"

Screenshot 2022-02-15 at 11-36-28 projen 0 52 31 - Construct Hub

### DIFF
--- a/src/views/Package/PackageDocs.tsx
+++ b/src/views/Package/PackageDocs.tsx
@@ -1,11 +1,5 @@
 import { Box, Flex, Grid } from "@chakra-ui/react";
-import type {
-  ClassSchema,
-  EnumSchema,
-  InterfaceSchema,
-  StructSchema,
-} from "jsii-docgen";
-import { FunctionComponent, useEffect, useMemo } from "react";
+import { FunctionComponent, useEffect } from "react";
 import {
   Route,
   Switch,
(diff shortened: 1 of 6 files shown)


src/views/Package/PackageDocs.tsx
```diff
@@ -48,171 +43,28 @@ const SubmoduleSelector: FunctionComponent = () => {
   ) : null;
 };
 
-const isApiPath = (path: string) => {
```
gabewomble, line 51: I found this component file a bit overwhelming when I jumped into the task, so I extracted most of the business logic here into it's own module `useSectionItems`. This allows the `PackageDocs` component file to focus on presentation logic

Chriscbr (approved):
🚢

PR #962, opened 2022-04-21
feat: add support for RSS/ATOM feed

yuth:
This change adds link tag in the frontend to support RSS/ATOM feeds for latest packages. 

re: https://github.com/cdklabs/construct-hub/pull/855/

### DIFF
--- a/src/api/config/index.ts
+++ b/src/api/config/index.ts
@@ -94,7 +94,7 @@ export interface FeaturedPackagesDetail {
 export interface FeatureFlags {}
 
 export interface Feed {
-  type: string;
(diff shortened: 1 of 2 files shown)


src/api/config/index.ts
```diff
@@ -93,13 +93,19 @@ export interface FeaturedPackagesDetail {
 
 export interface FeatureFlags {}
 
+export interface Feed {
+  type: string;
```
Chriscbr, line 97: Just double checking this is the right field (in the other PR we changed a field named type to mimeType)

Chriscbr (approved):

PR #997, opened 2022-06-11
feat: github badges

Chriscbr:
Closes https://github.com/cdklabs/construct-hub/issues/232

Corresponding backend PR: https://github.com/cdklabs/construct-hub/pull/911

Create GitHub badges that can be placed on repos to link users to a package's page on Construct Hub. A new tab is added to the package page which provides easy-to-copy snippets. The badge image will reside at `constructs.dev/badge`, ~and it's possible to pass to pass `darkMode=true` as a query parameter to get a dark badge instead.~

By default, the URLs provided by the front-end also include the package name as a `package` query parameter so that in the future, we could extend this to have the endpoint go to a lambda function which could generate a unique image based on the package name. For example, the badge could be extended to show what languages we have documentation for.

For the badge images to work, this also requires a separate PR on the backend to be merged (I think it should be merged first).

~I'm not sure the dark badge is necessary, but hey people like choice so why not give it to them.~ 😄

[Revision 1]: Updated the badge to be dynamically themed based on the user's browser client. I've still included assets for statically-colored buttons in the app in case anyone wants to use those.

Screenshot of the new tab added on the front-end:

Screen Shot 2022-06-13 at 10 36 20 AM

Screenshot of what this would look like on a GitHub README:

Screen Shot 2022-06-10 at 4 37 24 PM


RomainMuller (approved):
I think the badges are just a little too big but OTOH I think they're a good addition... the looks can be tuned later...  😅

PR #1059, opened 2022-08-14
fix: change `CDKtf` to `CDKTF`

### LINKED ISSUE #1058: CDKTYPE_NAME_MAP refers to TerraForm in lower case
Based on the TF team, they preference to have CDKTF in capital letters.

evgenyka:
Fixes #1058: change CDKTYPE_NAME_MAP [CDKType.cdktf]: "CDKTF" instead of "CDKtf"

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -6,7 +6,7 @@
 
   <meta charset="utf-8" />
   <meta name="description"
-    content="Construct Hub helps developers find open-source construct libraries for use with AWS CDK, CDK8s, CDKTf and other construct-based tools." />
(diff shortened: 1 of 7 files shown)


iliapolo (commented):
@evgenyka There are many other places in the code that use lowercase. For example:

https://github.com/cdklabs/construct-hub-webapp/blob/d219204d51fae18c482e2704bb903886054a96b8/src/views/Home/Home.tsx#L18-L25

Maybe run a global search so we can knock these out?

evgenyka:
Done - searched in all files and replaced CDKtf with CDKTF.

iliapolo (approved):

